PR #64, opened 2026-07-08
feat(config): edit ConfigMap and Secret values in place from the overview

deveshk0:
Adds in-place editing of ConfigMap and Secret values directly from the detail Overview page (requested). All test-first.

## Backend (`crates/kube`)
- **`k8s.updateConfigData`** — merge-patches ConfigMap `data` / Secret `stringData`. Using `stringData` means the apiserver base64-encodes, so **the UI only ever handles plaintext** and only the *changed* keys are sent (a merge patch leaves the rest untouched). `requires_confirm`; registered → auto-exposed as an MCP tool (completeness green).
- Pure `build_config_patch` unit-tested: `data` for ConfigMap, `stringData` for Secret, error for unsupported kinds.

## Frontend
- `lib/actions.updateConfigData`.
- **`ConfigDataEditor`** used by `ConfigBody` + `GeneralSecretBody`:
  - ConfigMap values are editable textareas; a **Save** (and Reset) appears only when something changed.
  - **Secret values stay masked until "Reveal"** (which decodes for editing) — never shown otherwise; the security property from the read-only view is preserved.
  - **Immutable Secrets render read-only** (no context passed → no Save).
  - Editing is only enabled where a `context` exists (the live detail); pure `ObjectDetail` previews stay read-only.
- After a successful save the overview **re-fetches** the object (local reload) so the panel and its summary reflect the new values.

## Verification
9 Rust + 289 frontend tests, coverage above gates, `tsc` clean, production build clean, MCP completeness green.

### DIFF
--- a/apps/desktop/src-tauri/src/capabilities.rs
+++ b/apps/desktop/src-tauri/src/capabilities.rs
@@ -91,6 +91,7 @@ pub fn build_registry_with(cache: Arc<ClientCache>) -> Registry {
     reg.register(srelens_kube::actions::delete_resource_capability(cache.clone()));
     reg.register(srelens_kube::actions::scale_capability(cache.clone()));
     reg.register(srelens_kube::actions::rollout_restart_capability(cache.clone()));
+    reg.register(srelens_kube::actions::update_config_data_capability(cache.clone()));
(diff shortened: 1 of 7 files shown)


navneet066 (approved):
LGTM

Issue #23, opened 2026-07-06
MCP security: consent toggles, loopback token, confirm policy


deveshk0:
The spec commits to consent-gated MCP: off by default, per-transport, destructive tools confirm-gated unless a session is granted autonomous scope.

## Scope
- [ ] Settings > Agents: enable/disable stdio and HTTP transports independently (both off by default for the GUI process; the explicit \`--mcp-*\` CLI flags remain opt-in by nature)
- [ ] HTTP transport: rotating bearer token generated per session, required on every request, shown once in settings for client config; still loopback-bound
- [ ] Policy control: destructive tools always require \`_confirm\` (default) vs session-granted autonomous mode with an explicit, visible opt-in and revoke
- [ ] Audit log view: recent MCP tool invocations (tool, args summary, outcome)

## Acceptance criteria
- A fresh install exposes nothing until the user opts in; HTTP requests without the token are rejected; the audit view shows every call

deveshk0:
PR #185 implements this — consent gating, bearer-token auth, and the audit log.

Two notes on scope, so the divergence from the issue text is on the record rather than buried in a diff:

**No GUI toggle for stdio.** The issue asks to "enable/disable stdio and HTTP transports independently". HTTP has a toggle. stdio does not, because a GUI cannot govern a process that an MCP client spawned — `srelens --mcp-stdio` runs with whatever flags the client's config gave it, and the app has no handle on it. The CLI flags (`--mcp-allow-destructive`, `--mcp-token`) are stdio's control surface, and this is documented in `docs/USAGE.md` rather than left for someone to hunt for a switch that cannot exist.

**Two things were found ungated while implementing this**, both now fixed in the PR:
- `k8s.helmRepoUpdate` carried `Annotations::default()` — neither read-only nor confirm-gated — so it executed over MCP with no consent step. The completeness assertion has been widened to fail the build on *any* capability that is neither read-only nor confirm-gated, so this class cannot recur.
- `k8s.getSecret` was not consent-gated either. `Annotations::SENSITIVE_READ` now implies confirmation.

Also fixed: rotating the token did not invalidate the old one while the server was running, which rather defeated the point of the button.

PR #66, opened 2026-07-08
feat(secrets): gated getSecret read, redact getObject, copy button + quota usage bars

deveshk0:
Closes the three items in #8 that were checked but not actually complete (per audit). test-first.

## Security: dedicated, policy-gateable Secret reads
Previously Secret **values** flowed through the generic `k8s.getObject` — an MCP client could read them with nothing to gate it. Now:
- New **`sensitive`** annotation flag (`Annotations::SENSITIVE_READ`) so a consent layer (#23) can gate secret-value reads separately from ordinary reads.
- **`k8s.getSecret`** capability (`sensitive`) — the only structured path to Secret values. Registered → MCP-exposed.
- **`k8s.getObject` now redacts Secret `data`** (keys kept, values blanked). The detail views fetch real values through `getSecret` (`SecretBody`'s `useSecretData`); values sit in memory but stay **masked in the DOM until the user reveals a key**. A backend unit test asserts redaction blanks values but keeps keys.

## UX
- **Copy button** per key on ConfigMap/Secret values (revealed values only).
- **ResourceQuota usage bars** — `parseQuantity` (handles `m` / `Ki…` / `k…` suffixes) drives a utilisation bar coloured green/amber/red, replacing the plain used/hard text.

## Verification
9 Rust + 299 frontend tests, coverage above gates, `tsc` clean, production build clean, MCP completeness green.

Note: rebased on dev after #62/#63/#64 merged; this branch predates the toasts PR #65 (independent).

### DIFF
--- a/apps/desktop/src-tauri/src/capabilities.rs
+++ b/apps/desktop/src-tauri/src/capabilities.rs
@@ -110,6 +110,7 @@ pub fn build_registry_with(cache: Arc<ClientCache>) -> Registry {
     reg.register(srelens_kube::nodes::list_nodes_capability(cache.clone()));
     reg.register(srelens_kube::manifest::get_manifest_capability(cache.clone()));
     reg.register(srelens_kube::manifest::get_object_capability(cache.clone()));
+    reg.register(srelens_kube::secrets::get_secret_capability(cache.clone()));
     reg.register(srelens_kube::manifest::apply_manifest_capability(cache.clone()));
     reg.register(srelens_kube::manifest::validate_manifest_capability(cache.clone()));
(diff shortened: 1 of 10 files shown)


navneet066 (commented):
This is a clear, well-scoped improvement. It removes Secret values from the generic structured path (k8s.getObject), introduces a gated, annotated k8s.getSecret for fetching Secret values, and updates the UI to fetch/mask/reveal values, add per-key copy actions, and show ResourceQuota usage bars. Tests cover the redaction and the frontend flows.

I approve this change — LGTM.

PR #97, opened 2026-07-09
feat: add functionality to remove clusters and automatically sync kubeconfig changes

shkatara:


deveshk0 (changes requested):
Reviewed end to end. Nice feature — real capability with an on-disk test, correctly `DESTRUCTIVE` (so agents need `_confirm`), cluster/user only removed when no longer referenced, `current-context` cleared when it's the one deleted, and thorough tab/profile cleanup plus external-change sync. A few things to address before merge, one of them a clear defect.

## 🟥 Critical

**1. `pnpm-workspace.yaml` contains placeholder junk**

```yaml
allowBuilds:
  esbuild: set this to true or false
```

`allowBuilds` isn't a valid pnpm-workspace key (the real one, `onlyBuiltDependencies: [esbuild]`, is already present two lines below), and `"set this to true or false"` is literal instructional text. This looks like an accidental commit — please remove it.

## 🟧 Important

**2. Kubeconfig rewrite is lossy and non-atomic** — `crates/kube/src/contexts.rs`

The handler does `serde_yaml::from_str` → mutate → `serde_yaml::to_string` → `std::fs::write` over the **whole** file. That silently drops **comments, key ordering, and anchors/aliases**, and a mid-write failure truncates the file. Kubeconfigs are frequently hand-maintained. Minimum: write atomically (temp file + `rename`). Ideally preserve unrelated content. At least call the comment-loss out in the PR/description.

**3. Native `alert()` / `window.confirm()`** — `App.tsx`, `SettingsView.tsx`

Use the app's `notify` toasts and `ConfirmDialog` instead of native dialogs — the native ones are inconsistent with the rest of the UI and render as blocking OS dialogs in the Tauri shell.

## 🟨 Minor

4. **Busy-poll watcher** (`lib.rs`) polls file mtime every 1.5s for the app lifetime; the `notify` crate would be event-driven. Also `cache.clear()` drops *all* clients on any change, and the tool's own write re-triggers the watcher (harmless double refresh).
5. **Backend test uses a fixed temp filename** (`srelens-test-delete-kubeconfig.yaml`) → parallel-run collision risk; use a unique path (e.g. include a nonce).
6. **Dual delete paths in `SettingsView`** (`onDeleteContext` vs the internal fallback) do *different* cleanup — the fallback skips the profile/namespace/tab cleanup the `App` handler does. Worth unifying so the fallback can't leave stale state.
7. Empty PR description; and the `deleteContext` lib wrapper doesn't follow the `{ error }` return convention the other wrappers use.

---

**Verdict: request changes** — the `pnpm-workspace.yaml` junk must go; the lossy/non-atomic kubeconfig write and the `alert`/`confirm` UX should be addressed before merge. The rest is solid.

deveshk0:
Pushed fixes for the review feedback (3 commits on top of `147999c`):

**`5882bfc` — startup crash (new, Critical).** The watcher was spawned with `tokio::spawn` inside Tauri's `setup`, which panics at launch (`there is no reactor running, must be called from the context of a Tokio 1.x runtime`). Switched to `tauri::async_runtime::spawn` so it runs on Tauri's managed runtime.

**`93f6732` — backend + config.**
- Removed the accidental `allowBuilds: esbuild: set this to true or false` placeholder from `pnpm-workspace.yaml` (Critical).
- Made the kubeconfig write **atomic** (temp file + `rename`, preserving unix perms) so a mid-write failure can't truncate the user's kubeconfig; documented that the serde_yaml round-trip doesn't preserve comments (Important).
- Backend delete test now uses a unique temp path (pid + nanos) to avoid parallel-run collisions (Minor).

**`d53f4f0` — frontend.**
- Replaced native `alert()` / `window.confirm()` with the app's `notify` toasts and `ConfirmDialog` (Important).
- Unified the two delete paths in `SettingsView` behind a single confirm gate + shared error handling (Minor).
- Added a **right-click context menu** on each context row in Settings → Contexts ("Reset identity" / "Remove context"), per request. Tests updated + a new right-click test added.

Verified: `cargo test -p srelens-kube -p srelens-desktop` green (121 tests), desktop crate compiles, `pnpm test` 411 passing with the 80% line-coverage gate met.

Left as-is by decision: the kubeconfig watcher stays on the 1.5s mtime poll rather than the `notify` crate — `notify` is fragile with the atomic temp+rename save pattern (needs directory watching + rename de-dup), and polling is simpler/robust for infrequent kubeconfig changes.

Note (pre-existing, not from this PR): the workspace `cargo llvm-cov --fail-under-lines 55` gate is under threshold at this branch's merge-base too (~53%), so it's independent of these changes.

deveshk0 (approved):

PR #98, opened 2026-07-09
feat(terminal): local kubectl terminal in the bottom dock (#96)

### LINKED ISSUE #96: Local kubectl terminal: open a shell tab in the bottom dock, scoped to the current context
## Problem

There's no way to run `kubectl` (or `helm`, etc.) from inside the app. The only terminal today is **pod exec** — `PodTerminal` → `start_pod_exec`, which runs a shell *inside a pod's container* via kube-rs. There is no **local** shell, so users must leave the app and switch contexts by hand to run ad-hoc commands.

## Request

Add a **terminal tab that opens from the bottom** (the existing Dock) running a **local shell**, pre-scoped to the **current Kubernetes context** — so `kubectl get pods` "just works" against the cluster the user is looking at, without `--context` juggling.

## Current building blocks

- **Dock** (`apps/desktop/src/components/Dock.tsx`) — the bottom panel already hosts sessions; `DockKind = "terminal" | "logs"` (today "terminal" == pod exec).
- **xterm** rendering is already wired in `PodTerminal.tsx` (`@xterm/xterm` + fit addon) — reusable for a local terminal.
- **Streaming bridge** — `exec.rs` streams a kube-rs exec session's stdout to the WebView over Tauri events with stdin back; the same event pattern fits a local PTY.
- `main.rs` already notes the app inherits the shell `PATH` so kubeconfig exec-plugins resolve — good, `kubectl` on PATH will be found.

## Gap / what's new

There is **no local process/PTY** anywhere — all exec is in-cluster. This needs:

1. **Backend — local PTY.** Spawn the user's `$SHELL` (fallback `/bin/bash`) in a pseudo-terminal (e.g. the `portable-pty` crate), stream stdout↔stdin over Tauri events like the exec bridge, and handle resize. Desktop-only (not exposed via MCP).
2. **Context scoping.** Point the child's `KUBECONFIG` at a generated temp kubeconfig — merged from the app's configured kubeconfig files (`loadKubeconfigFiles` + default) with `current-context` pinned to the active context — so plain `kubectl`/`helm` target that cluster. The terminal is **bound to the context it was opened with**.
3. **Frontend — a local terminal tab.** Add a Dock session kind (e.g. `shell`) reusing the xterm view; label the tab with the context name. Provide an entry point: a "Open terminal" action (toolbar button + command palette), opening the tab for the current cluster.

## Acceptance criteria

- [ ] A terminal tab opens in the bottom dock and runs a real local shell (interactive, resizable)
- [ ] `kubectl` in that terminal targets the current context by default (no `--context` needed)
- [ ] The tab shows which context it's bound to; opening from a different cluster gives a terminal for that cluster
- [ ] `kubectl`/`helm` binaries resolve via the user's PATH; a clear message if `kubectl` isn't installed
- [ ] Closing the tab terminates the child process cleanly (no orphaned shells)
- [ ] Local terminal is desktop-only and never exposed as an MCP tool

## Security / notes

- This runs **arbitrary local commands on the user's machine** (their shell, their kubectl) — expected for a kubectl terminal, but it is *not* sandboxed. Keep it strictly a local desktop feature; do not surface it through the MCP capability registry.
- Open question: full shell vs. a kubectl-restricted prompt. A full local shell is the most useful (helm, jq, pipes) and matches Lens; leaning that way unless we want to restrict.
- Follow-up ideas (out of scope): remember the working namespace, seed the prompt with the context name, per-tab context switching.

navneet066:
Closes #96

## What

An in-app **local terminal** — a real shell (your `$SHELL`) opened from the status bar and hosted in the bottom **dock** — so you can run `kubectl`/`helm` without leaving the app. Distinct from the in-pod exec terminal; **desktop-only, never exposed via MCP**.

## Backend

- **Local PTY bridge** (`portable-pty`) streaming stdout/stdin/resize over Tauri events: `start_terminal` / `terminal_input` / `terminal_resize` / `terminal_close`, keyed per session.
- **Locked to one context**: each terminal gets a standalone temp kubeconfig containing **only** the opened context (+ its cluster/user), written `0600`, set as `KUBECONFIG`. So `kubectl` targets it by default, `config get-contexts` lists just that one, and `use-context` can't switch away. (`srelens_kube::connect::single_context_kubeconfig_yaml`)
- Parent env (PATH resolved by `fix-path-env`) is propagated so `kubectl`/`helm` resolve.

## Frontend

- `LocalTerminal` (xterm) + `lib/terminal.ts` client.
- **Dock** gains a `shell` session kind, a **"+"** button to open another terminal for the same context, and now **keeps every session mounted** (hiding inactive panes) — switching tabs no longer tears down shells or loses scrollback.
- Tab bar **scrolls horizontally with truncated labels** (+ tooltip) so many long-named terminals don't break the layout.
- **StatusBar "Terminal"** launcher for the active cluster.

## Testing

- Rust: `single_context_kubeconfig_yaml` test (only the named context/cluster/user survive; missing context errors).
- Frontend: Dock tests (persistence/hide-not-unmount, "+" button) + StatusBar launcher test.
- Full suite green (**411**); `vite build` + `cargo check` clean.
- Verified live: locked context, multi-tab, tab persistence, layout.

## Notes / scope

- It's a **real shell on the user's machine** (their `$SHELL`/kubectl) — inherent that a determined user could `export KUBECONFIG=…` to escape the lock; the default/listed/switchable scope is the one cluster, which is the intent.
- UTF-8 output is forwarded lossily at chunk boundaries (fine for typical CLI output).
- Follow-up ideas: numbering same-context tabs to tell them apart; a command-palette entry.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -906,6 +906,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "downcast-rs"
+version = "1.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "75b325c5dbd37f80359721ad39aca5a29fb04c89279657cffdda8736d0c0b9d2"
+
+[[package]]
 name = "dpi"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -985,7 +991,7 @@ dependencies = [
  "rustc_version",
  "toml 1.1.2+spec-1.1.0",
  "vswhom",
- "winreg",
+ "winreg 0.55.0",
 ]
 
 [[package]]
@@ -1102,8 +1108,19 @@ version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "38e2275cc4e4fc009b0669731a1e5ab7ebf11f469eaede2bab9309a5b4d6057f"
 dependencies = [
- "memoffset",
+ "memoffset 0.9.1",
  "rustc_version",
+]
+
+[[package]]
+name = "filedescriptor"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e40758ed24c9b2eeb76c35fb0aebc66c626084edd827e07e1552279814c6682d"
+dependencies = [
+ "libc",
+ "thiserror 1.0.69",
+ "winapi",
 ]
 
 [[package]]
@@ -1993,6 +2010,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "ioctl-rs"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f7970510895cee30b3e9128319f2cefd4bde883a39f38baa279567ba3a7eb97d"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "ipnet"
 version = "2.12.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2412,6 +2438,15 @@ checksum = "88904434abc2901f197fe8cc55f0445e7ded921dba5911dad2e2b39b48e663c4"
 
 [[package]]
 name = "memoffset"
+version = "0.6.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5aa361d4faea93603064a027415f07bd8e1d5c88c9fbf68bf56a285428fd79ce"
+dependencies = [
+ "autocfg",
+]
+
+[[package]]
+name = "memoffset"
 version = "0.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "488016bfae457b036d996092f6cb448677611ce4449e970ceaf42695203f218a"
@@ -2502,6 +2537,20 @@ name = "new_debug_unreachable"
 version = "1.0.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "650eef8c711430f1a879fdd01d4745a7deea475becfb90269c06775983bbf086"
+
+[[package]]
+name = "nix"
+version = "0.25.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f346ff70e7dbfd675fe90590b92d59ef2de15a8779ae305ebcbfd3f0caf59be4"
+dependencies = [
+ "autocfg",
+ "bitflags 1.3.2",
+ "cfg-if",
+ "libc",
+ "memoffset 0.6.5",
+ "pin-utils",
+]
 
 [[package]]
 name = "num-conv"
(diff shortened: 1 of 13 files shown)


deveshk0:
Reviewed and pushed fixes for the issues found (`289a94c`). Nice feature overall — the single-context kubeconfig isolation is clean and well-tested, and it's correctly desktop-only / never MCP-exposed.

**Fixed (Important):**

1. **Insecure temp kubeconfig** — `write_locked_kubeconfig` wrote a predictable `/tmp/srelens-term-<id>.kubeconfig` with `fs::write` then `chmod 0600`. That's symlink/TOCTOU-prone (a pre-existing symlink at the path would be followed and clobbered) and leaves a world-readable window before the mode is applied. Now created atomically with `OpenOptions::create_new(true)` (O_CREAT|O_EXCL, so an existing path/symlink can't be followed or overwritten) + `mode(0o600)` at creation, and the filename includes `process::id()` for cross-run uniqueness.

2. **Backend session leak on natural exit** — when the shell exited on its own, the `Session` was never removed from `TerminalManager.sessions` (the reader thread had no handle to the map), leaking the PTY master/child for the app's lifetime. `sessions` is now `Arc<Mutex<…>>` and the reader thread reaps its own entry (`remove(&id)`) when it hits EOF. The frontend `onExit` also calls `session.close()` now (tears down listeners + `terminal_close`).

**Fixed (Minor):**
- The `connect.rs` single-context test used a fixed temp filename → now unique (pid + nanos) to avoid parallel-run collisions.

**Left as acknowledged Minors (not blocking):**
- Early-output race (events emitted before the frontend subscribes) — rare in practice since shell startup > the IPC round-trip; can be closed later with a client-provided channel token if it shows up.
- UTF-8 forwarded lossily at 8 KiB chunk boundaries (already noted in the PR).
- Blocking `std::fs` in the async `start_terminal` — negligible for a one-time small write on terminal open.

Verified: `cargo test -p srelens-kube` (115) + `every_capability_is_mcp_exposed` green, desktop crate compiles clean, `pnpm test` 411 passing.

deveshk0:
Follow-up (`c1798a4`): also fixed the three items I'd initially flagged as acknowledged Minors —

- **Early-output race** — the frontend now generates a channel token and `subscribe()`s (awaiting registration) *before* `start_terminal` spawns, so the shell's first prompt can't race ahead of the listener. Backend emits on `term:out:<channel>` / `term:exit:<channel>`; the numeric session id still drives input/resize/close.
- **UTF-8 chunk boundaries** — the reader thread now carries an incomplete trailing multibyte sequence between 8 KiB reads (flushing >3-byte leftovers lossily so it can't stall), instead of `from_utf8_lossy` per chunk.
- **Blocking fs** — the kubeconfig create/write runs on `tokio::task::spawn_blocking` off the async runtime.

Verified: `cargo check`/`cargo test` green, `pnpm test` 411 passing. Also confirmed this branch still merges cleanly on top of the merged #97 (no conflicts in `lib.rs`/`App.tsx`).

Issue #111, opened 2026-07-12
AppImage shows blank/white window on Wayland due to bundled libwayland-* libs conflicting with system Mesa (EGL_BAD_PARAMETER)


kumailr7:
### Summary
The `srelens_0.2.0_amd64.AppImage` shows a completely blank/white window on launch, with no visible content. The only diagnostic output is:

```
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

Setting `WEBKIT_DISABLE_DMABUF_RENDERER=1` (a commonly-suggested WebKitGTK workaround) does **not** fix it.

### Root cause
The AppImage bundles its own copies of:

```
usr/lib/libwayland-egl.so.1
usr/lib/libwayland-client.so.0
usr/lib/libwayland-cursor.so.0
usr/lib/libwayland-server.so.0
```

These bundled libs are older than the system's Mesa/Wayland stack and are incompatible with it, causing WebKitGTK's `eglGetDisplay()` call to fail with `EGL_BAD_PARAMETER` when running under a Wayland session (confirmed via XWayland/X11 too). Since Wayland/EGL libraries are effectively part of the platform ABI (they must match the compositor/driver stack, similar to Mesa/graphics libs), vendoring them in the AppImage causes this class of failure on any system with a newer Mesa than the one used at build time.

### Fix that resolved it locally
Removing the four bundled `libwayland-*` files above from the AppImage's `usr/lib/` (after `--appimage-extract`), so the app falls back to the system's Wayland libraries via the dynamic linker, completely resolves the issue — no more EGL error, window renders correctly.

### Suggested fix
Exclude `libwayland-egl`, `libwayland-client`, `libwayland-cursor`, and `libwayland-server` from the `linuxdeploy`/AppImage bundling step (e.g. via `linuxdeploy`'s exclude-list) so the app always links against the host's Wayland/EGL libraries at runtime, similar to how libGL/EGL/GPU driver libs are typically excluded from AppImages.

### Environment
- Distro: Arch Linux (CachyOS), kernel 7.1.3-1-cachyos
- Desktop Environment: GNOME
- Session type: Wayland
- GPU: Intel Iris Xe Graphics (TigerLake-LP GT2), `i915` driver
- Mesa: 26.1.4-arch3.1
- srelens version: 0.2.0 (amd64 AppImage)

### Screenshot

[...]

PR #112, opened 2026-07-14
feat(packaging): AUR package (srelens-bin) for Arch/CachyOS

deveshk0:
Adds an AUR package (`srelens-bin`) so Arch / CachyOS / Manjaro / EndeavourOS users can `paru -S srelens-bin`.

## Why

The AppImage vendors `libwayland-{client,cursor,egl,server}` and loads them ahead of the system's. On a rolling distro with a much newer Mesa, the host's EGL resolves against those stale bundled libs, `eglGetDisplay()` fails with `EGL_BAD_PARAMETER`, and srelens opens a **blank window with no error** (#111 — first reported on CachyOS + GNOME).

I confirmed this by extracting our actually-shipped AppImage: those four libs **are** bundled, while `libEGL`/`libGL` are **not** — a **half-vendored graphics stack**. That's the bug.

A `-bin` package links the system's own `webkit2gtk-4.1` / `gtk3` / wayland / Mesa, so that entire class of failure is *structurally impossible* rather than patched around.

## The package

- **Dependencies derived from the binary's real `DT_NEEDED`**, not guessed: `webkit2gtk-4.1` (supplies javascriptcore + libsoup3) and `gtk3`. Note the binary **never links `libwayland` directly** — it arrives transitively via GTK/Mesa, which is precisely why the AppImage bundling it was gratuitous.
- **`kubectl` / `helm` are `optdepends`, not `depends`** — the honest encoding of srelens's design: it drives whatever toolchain is already on your machine rather than bundling one.
- Ships the MIT `LICENSE` (Arch requires it) and fills in the **empty `Categories=`** our generated `.desktop` currently has, which can hide the app from desktop menus. Worth fixing upstream in `tauri.conf.json` too.

## Publishing

Runs in an `archlinux:base-devel` container on **stable releases only** (AUR expects released versions, and an Arch `pkgver` cannot carry the `-<run>` suffix our dev pre-release tags use). It no-ops until the AUR secrets exist, so it can land today.

Two deliberate choices:

- **No third-party action.** An AUR PKGBUILD **executes arbitrary code on a user's machine at build time**, so push access to this package is a high-value target. Handing the AUR SSH key to an external action would put that in someone else's supply chain to save ~20 lines of shell. We push with plain `git`/`ssh`.
- **It builds and `namcap`-lints the package before pushing** — a PKGBUILD that doesn't build must never reach users.

## Verification

Ran the exact CI sequence in a real Arch container: `updpkgsums` → `.SRCINFO` → `makepkg` → `srelens-bin-0.2.1-1-x86_64.pkg.tar.zst` produced, with the packaged `.desktop` carrying `Categories=Development;`.

## Before it can actually publish (needs a human)

1. **A stable release.** Every release so far is a rolling `dev` pre-release; an Arch `pkgver` can't carry `-<run>`. Needs a stable `srelens-v<version>` from `main`.
2. **Claim `srelens-bin` on the AUR** (the name is free) with one manual push, then add `AUR_USERNAME` / `AUR_EMAIL` / `AUR_SSH_PRIVATE_KEY`. Steps in `packaging/aur/README.md`.

Related to #111 — this gives Arch users a working channel, but does **not** by itself fix the AppImage for everyone else.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -365,3 +365,98 @@ jobs:
           if ls *.asc >/dev/null 2>&1; then
             gh release upload "$TAG" --repo "${{ github.repository }}" *.asc --clobber
           fi
+
+  # Arch / CachyOS / Manjaro install path. Deliberately STABLE-ONLY: AUR users
+  # expect released versions, and an Arch `pkgver` cannot carry the `-<run>`
+  # suffix that dev pre-release tags use.
+  #
+  # This also sidesteps a real bug class: the AppImage vendors libwayland-*, which
+  # the host's newer Mesa then resolves against, breaking EGL and yielding a blank
+  # window. A -bin package linking the SYSTEM webkit2gtk/gtk3/wayland/mesa cannot
+  # hit that. See packaging/aur/README.md.
+  #
+  # Requires (one-time, see packaging/aur/README.md): the srelens-bin name claimed
+  # on the AUR, plus AUR_USERNAME / AUR_EMAIL / AUR_SSH_PRIVATE_KEY secrets. The
+  # job no-ops when the key isn't configured, so it can land before that's done.
+  aur:
+    name: publish AUR (srelens-bin)
+    needs: [version, build]
+    if: ${{ github.ref_name == 'main' && needs.build.result == 'success' && needs.version.outputs.release == 'true' }}
+    runs-on: ubuntu-22.04
+    # Do the AUR push ourselves rather than via a third-party action: an AUR
+    # PKGBUILD executes arbitrary code on a user's machine at build time, so
+    # push access to the package is a high-value target. Handing the AUR SSH key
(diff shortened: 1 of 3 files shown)


deveshk0:
**Correction:** I previously said this was blocked on a first stable release — that's wrong. **`srelens-v0.2.0` already exists** as a stable (non-prerelease, `latest`) release with a published `.deb`. My earlier check only listed the most recent releases, which were all rolling `dev` pre-releases.

So the AUR package can be imported **today**. Verified: the unmodified PKGBUILD with `pkgver=0.2.0` (the only change CI makes) downloads the real v0.2.0 `.deb`, pins real checksums, builds, and produces `srelens-bin-0.2.0-1-x86_64.pkg.tar.zst` with `Categories=Development;` and the licence installed.

The only remaining prerequisite is the human step: claiming `srelens-bin` on the AUR and adding the `AUR_*` secrets.

PR #114, opened 2026-07-19
fix(release): keep libwayland-* out of the AppImage

### LINKED ISSUE #111: AppImage shows blank/white window on Wayland due to bundled libwayland-* libs conflicting with system Mesa (EGL_BAD_PARAMETER)

deveshk0:
Closes #111.

## Problem

The AppImage bundles the build machine's `libwayland-{client,cursor,egl,server}`. These are platform-ABI libraries — they must match the host's compositor/Mesa stack — so on any host with a newer Mesa (the reporter: CachyOS, Mesa 26), `eglGetDisplay()` fails with `EGL_BAD_PARAMETER` and srelens opens a blank window. The reporter confirmed that deleting the four bundled libs fixes it.

## Why they get bundled

- tauri-bundler (CLI 2.11.3) drives `linuxdeploy` with no exclusion flags and no config hook to add any.
- The linuxdeploy build tauri pins is from **2024-07**. `libwayland-client.so.0` joined the upstream excludelist in **2024-11** ("New version of Mesa has some dependency issues with libwayland-client if it is bundled"), and the other three libs aren't excluded even by current builds — verified empirically.
- linuxdeploy honors `LINUXDEPLOY_EXCLUDED_LIBRARIES` (glob patterns, appended to the built-in excludelist) — but only since **2025-08**, so the pinned binary is too old for it.

## Fix (all in `release.yml`)

1. **Seed the tools cache** (Linux job): download a current linuxdeploy into `~/.cache/tauri/` before the build — tauri-bundler uses a pre-existing cached binary instead of downloading its pin. Checksum-pinned; if upstream's rolling release changes, the step fails with an actionable message rather than silently switching binaries.
2. **`LINUXDEPLOY_EXCLUDED_LIBRARIES: "libwayland-*"`** on the build step — covers all four libs.
3. **Regression tripwire**: after the build, extract the AppImage and fail the job if any `libwayland-*` is inside. `updater-manifest` requires build success, so a regressed AppImage can never be offered to existing users via self-update.

## Verification

- Mechanism proven in an amd64 container replicating the bundler's exact invocation (including its `dd` magic-byte patch on the cached binary): a test binary linking all four wayland libs → default deploy bundles `cursor`/`egl`/`server`; with the env var, none are deployed.
- Bundler pickup of the seeded binary verified against tauri-bundler 2.11.3 source: exact cache path + filename match, `useLocalToolsDir` unset, `XDG_CACHE_HOME` not overridden in the workflow.
- Final on-artifact proof comes from the first CI run of this workflow: the tripwire step asserts the built AppImage is clean before the manifest can publish.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -117,6 +117,20 @@ jobs:
       # Apple secret set is present; with anything less the build stays
       # unsigned rather than failing on a partial configuration.
       ENABLE_MAC_SIGNING: ${{ secrets.APPLE_CERTIFICATE != '' && secrets.APPLE_CERTIFICATE_PASSWORD != '' && secrets.APPLE_SIGNING_IDENTITY != '' }}
+      # Libraries the AppImage must NOT bundle (#111) — see the seed step below.
+      # Single source of truth: the build passes this to linuxdeploy, and the
+      # post-build assert greps for the same pattern. Keep them in lockstep.
+      APPIMAGE_EXCLUDED_LIBS: "libwayland-*"
+      # Immutable linuxdeploy release (NOT the rolling `continuous` tag, whose
+      # asset is rebuilt by a monthly upstream cron and would break this pin —
+      # and every release with it — without a single upstream commit).
+      #
+      # Bumping this adopts upstream's baked-in excludelist wholesale, which can
+      # silently stop bundling libraries beyond the ones named here (it already
+      # covers libwayland-client and libpipewire-0.3). Diff that list before
+      # moving the pin; the assert below only guards the Wayland case.
+      LINUXDEPLOY_RELEASE: "1-alpha-20251107-1"
+      LINUXDEPLOY_SHA256: "c20cd71e3a4e3b80c3483cef793cda3f4e990aca14014d23c544ca3ce1270b4d"
     strategy:
       fail-fast: false
       matrix:
@@ -143,7 +157,7 @@ jobs:
 
       # Tauri v2 build dependencies.
       - name: Install Linux dependencies
-        if: matrix.platform == 'ubuntu-22.04'
+        if: runner.os == 'Linux'
         run: |
           sudo apt-get update
           sudo apt-get install -y \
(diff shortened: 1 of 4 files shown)


deveshk0:
Pushed `7ea45b0` addressing an independent review of the original commit. Two issues were significant enough to change the design:

**1. The pin would have broken every release, on a schedule.** The original pinned linuxdeploy's rolling `continuous` asset. Upstream rebuilds it on a **monthly cron** (`schedule: '0 0 1 * *'`, "build at least once a month") independent of commits, and the builds aren't reproducible — the current asset is stamped 2026-07-01 with no commits since May. The checksum pin would have failed by Aug 1, and since the seed step gates the Linux leg and `updater-manifest` requires all legs, that blocks **every release on every platform** until a human rotates the hash.

Fixed by pinning the immutable `1-alpha-20251107-1` release instead (post-dates the 2025-08 env-var feature). Re-ran the container verification against it: all four libs logged as `Skipping deployment of blacklisted library`. No mirror infrastructure needed.

**2. The assert ran after the release was already public.** `tauri-action` published a non-draft release before the tripwire, so on `main` a later failure left a release that `releases/latest` resolves to but with no `latest.json` attached — a 404 on the stable update check for the entire install base until manually fixed, plus the bad AppImage publicly downloadable. This failure class pre-dated the PR; the PR added new triggers for it.

Fixed by drafting the stable release and adding a `publish-release` job that flips the draft only after the manifest and signatures land (`aur` now waits on it, since `updpkgsums` needs public URLs). Dev deliberately stays non-draft — prereleases never resolve as `latest`, and drafting there would briefly point `dev-channel/latest.json` at 404ing draft assets.

**Also addressed:** single source of truth for the exclusion glob so the build and the assert can't drift; Linux steps keyed off `runner.os` so a runner bump can't silently skip the tripwire that guards the fix; loud failure on anything but exactly one built AppImage; dropped error handling that duplicated step-failure semantics; a note that bumping the pin adopts upstream's excludelist wholesale (the assert only guards the Wayland case).

**Docs:** the AUR README, PKGBUILD, and the `aur` job comment all asserted in the present tense that the AppImage bundles `libwayland-*` — false once this lands, so they now scope that to releases up to 0.2.1. `docs/INSTALL.md` documents the flip side: the AppImage now relies on the host's Wayland runtime (`libgdk-3`/`libwebkit2gtk` link it directly), which every GTK desktop has but a stripped-down image may not.

The tripwire's shell logic was tested against all four edge cases (zero/one/two AppImages, libs present/absent).

PR #165, opened 2026-07-28
feat: web mode (multi-user Docker webapp) + managed cluster OIDC sign-in

deveshk0:
## Summary

This branch does two big things:

1. **Runs srelens as a multi-user web app in a Docker container** — a headless server that serves the built frontend, authenticates users via OIDC (or a local dev login), and gives each user a fully isolated environment built only from their own uploaded, encrypted kubeconfigs. The desktop app is unchanged.
2. **Adds managed cluster OIDC sign-in (Headlamp-style)** on **both** web and desktop — srelens runs the OIDC authorization-code + PKCE flow in the browser itself and injects the id_token as the Bearer, so OIDC-protected clusters work **without a `kubelogin`/exec plugin** installed. On the web container that plugin can't run at all; on desktop it means one-click sign-in with no plugin required.

## Web mode

- New Tauri-free `srelens-server` binary (axum + sqlx/SQLite, embedded frontend via rust-embed) and a `crates/registry` split so the capability assembly is shared by desktop, the web server, and MCP.
- **Per-user isolation**: each user's capability calls, watches, and streams run against a registry + client cache built only from their own kubeconfigs, materialized as 0600 files under `<data>/runtime/users/<id>/` (wiped at startup).
- **App auth**: OIDC (authorization-code + PKCE + browser-binder cookie) or a local dev login; session cookies; CSRF via a required header + Origin checks for WebSockets; a per-user reverse proxy for port-forwards.
- **Streaming** (watch / logs / exec / port-forward / terminal / helm) refactored behind an `EventSink` trait (`crates/streams`) and multiplexed over a single `/api/ws`.
- **Docker**: multi-stage image (node build → rust build → debian-slim runtime, non-root, checksum-verified kubectl/helm), `docker-compose.yml`, health/readiness probes, jemalloc + capped worker threads for memory. `docs/WEB.md` covers deployment, env vars, and security requirements.

## Cluster OIDC (web + desktop)

- Detects OIDC contexts from a kubeconfig's `auth-provider: oidc` **and** `exec` kubelogin/`kubectl oidc-login` forms; groups contexts by issuer+client so one sign-in serves many.
- **Web**: `/auth/cluster/login|callback` routes run the browser flow; tokens are sealed per (user, cluster); the resolver injects the Bearer (stripping the exec block) or returns a `cluster_login_required` 401 that surfaces a lazy "Sign in to `<cluster>`" prompt. An "Add cluster" form and per-cluster sign-out are in Settings → Contexts.
- **Desktop**: the same managed flow via a loopback (`127.0.0.1`) callback (RFC 8252, like `kubelogin`), a local sealed sqlite token store in the app config dir, and the resolver installed on the desktop client cache. Reuses the server's OIDC machinery directly — no reimplementation. Non-OIDC exec plugins (aws/gke) still run natively on desktop.
- A "Test connection" button on both add paths validates reachability without running any exec plugin (RCE-safe on the shared server).

## Notes

- **Desktop is additive**: the OIDC resolver returns "default" for non-OIDC contexts, so native auth is byte-for-byte unchanged; all new web-mode UI is gated off `isTauri()`.
- Full suite green: server, kube, streams, registry Rust tests + the frontend vitest suite; TLS stays rustls (no aws-lc/openssl).

## Known limitations / follow-ups (documented, non-blocking)

- Reactive token refresh on a mid-session 401 (currently proactive-only).
- Private-CA IdPs, Azure `kubelogin`, and non-OIDC cloud plugins (aws/gke) — see `docs/WEB.md`.
- Web terminal/helm can't use the managed OIDC token (CLI subprocess limitation).


deveshk0:
## Review findings → fixes

Two independent security reviews (isolation/auth and OIDC/desktop) plus CodeQL. Mapping each finding to the commit that resolves it.

| # | Finding | Sev | Status | Commit |
|---|---------|-----|--------|--------|
| 1 | `start_terminal` on `/api/command` spawns a host shell as the shared UID (no web deny-list) → cross-user RCE, reads other users' materialized kubeconfigs + sealed tokens | 🔴 Critical | Fixed | `96bcda6` |
| 2 | `start_helm_op` runs `helm <raw args>`, bypassing the capability deny-list → `repo add` poisoning + `plugin install <url>` RCE | 🟠 High | Fixed | `96bcda6` |
| 3 | macOS loopback OIDC callback dropped (accepted socket inherits `O_NONBLOCK` → `read()` WouldBlock) → sign-in hangs | 🟡 Medium | Fixed | `8bd8960` |
| 4 | CodeQL: client-side redirect / FakeIdp fixed nonce | — | Fixed earlier | `16b63f2` |

### How 1 & 2 were resolved
Rather than sandbox the host shell, the web surface now refuses the two paths that have **no safe multi-user form**, and keeps the equivalents srelens already had (this matches how Headlamp works — RBAC-scoped in-pod exec, structured/isolated helm):

- **Terminal** — `start_terminal` returns `403` in web mode and its UI entry points are hidden. Web users get the RBAC-scoped in-pod `start_pod_exec` terminal. Host shell stays desktop-only.
- **Helm** — the web surface rejects `repo`/`plugin` subcommands (and leading-flag evasion), and every helm run is scoped to a **per-user HELM home** (`HELM_CONFIG_HOME`/`HELM_CACHE_HOME`/`HELM_DATA_HOME`). install/upgrade/rollback/uninstall stay available. Desktop unchanged. (There is no native Rust Helm SDK — the Go SDK Headlamp uses isn't available to a Rust backend — so the fix sandboxes our `helm` subprocess instead.)

### Additional hardening from review discussion (`e133d9c`)
- **Master key must come from the environment.** Server mode previously auto-generated `master.key` next to the encrypted DB (volume held ciphertext *and* key). It now requires `SRELENS_MASTER_KEY` and fails fast if missing; the key is never written to the data volume.
- **Decrypted material off the durable disk.** `kubectl`/`helm`/kube-rs need real files, so per-user kubeconfigs + helm state are materialized under `/data/runtime`; `docker-compose.yml` now backs that path with **tmpfs** (RAM, 0700, container uid/gid).

Docs updated in `85a478f` (WEB.md) and `b50b2a5` (README).

### Verified sound by the reviews (not changed)
Per-user env isolation, session/CSRF + Origin checks, OIDC login binder+state, cluster-OIDC tokens scoped per user with no cross-user access, token-aware client cache (bearer-keyed, exec stripped), port-forward proxy ownership + cookie stripping, WebSocket user-scoping, RCE-safe test-connection probe, and the loopback CSRF/injection model (RFC 8252 residual only).

### Open, non-blocking (tracked as fast-follows)
OIDC nonce generated but not verified; duplicate context names lose *managed* OIDC (desktop falls back to native, web shows guidance); stale refresh token replay after a single-flight wait; OIDC init failure leaves Settings commands on unmanaged state; loopback success page always says "Signed in". Low severity — proposed as follow-ups.

Ready for re-review.

Issue #172, opened 2026-07-31
test(forward): live-cluster e2e for port-forward pod-deletion recovery


deveshk0:
Follow-up from #21. The auto-reconnect logic (backend detects target-pod loss → re-resolves the service to a replacement pod → reconnects) is unit-tested at the pure-helper level (`pod_is_gone`, `next_reconnect_state`) and verified sound by code review, but the **full acceptance path has no automated regression test** because it needs a live kube apiserver (or a fake) that the current unit-test infra doesn't provide.

Add a `kind`-based e2e (in `apps/desktop/src-tauri/tests/e2e.rs`, which already runs against a kind cluster): start a **service** port-forward, delete the backing pod, and assert the forward reports `reconnecting` then `active` again once a replacement pod is Ready — matching #21's acceptance criterion.

PR #173, opened 2026-07-31
feat(forward): resilient port-forwards + saved forwards

### LINKED ISSUE #21: Port-forward: auto-reconnect, conflict handling, saved forwards
## Scope
- [ ] Auto-reconnect with backoff when the backing pod restarts (re-resolve target from service/workload selector)
- [ ] Local port conflict detection with suggested free port
- [ ] Saved forwards per cluster (name, target, ports) that can be started with one click on connect
- [ ] Forward status surfaced (active/reconnecting/failed) in the indicator and the forwards view

## Acceptance criteria
- Deleting the target pod of an active service forward recovers without user action once a replacement pod is ready

deveshk0:
Closes #21.

Makes port-forwards survive pod restarts, detects local-port conflicts, and adds named saved-forwards per cluster with live status.

## What's in it
- **Auto-reconnect** (`crates/streams/forward.rs` + `crates/kube/forward.rs`) — when a forward drops, the backend retries with capped exponential backoff and, for service/workload forwards, **re-resolves to a fresh pod each attempt**, then gives up after N consecutive failures. A target-pod liveness monitor ends the serve session on pod deletion so recovery actually fires (the acceptance criterion). Consecutive-failure counting resets on each successful establish.
- **Status channel** — emits `forward:status:<id>` (`active`/`reconnecting`/`failed`), surfaced in `ForwardsIndicator` and `PortForwardsView`.
- **Local-port conflict detection** — `bind_local` returns a typed `BindError::InUse { requested, suggested }`; `ForwardDialog` offers a one-click retry on the suggested free port.
- **Saved forwards per cluster** — a `lib/savedForwards.ts` store: **web** persists to a new per-user `GET/PUT/DELETE /api/settings/:key` slice (previously-unused `settings` table); **desktop** uses `localStorage`. Managed from the Saved section of `PortForwardsView` (one-click Start / Delete), with errors surfaced (not silent).

## Tests
Backend: `bind_local` conflict, `pod_is_gone`, `next_reconnect_state` transition table, the retry→failed loop (streams 21, kube 282); per-user settings DB + route tests (server 130). Frontend: savedForwards store round-trips, status-event wiring, saved-forwards UI (Start/Delete/error), status rendering, suggested-port retry (**716/716**, tsc clean). No AI attribution across the 8 commits.

## Built via subagent-driven development
5 tasks, per-task reviews (spec + quality) with fix loops; a Critical (pod-deletion didn't trigger re-resolve) and several Important findings were caught and fixed in review.

## Known follow-ups (not blocking)
- **Live-cluster e2e** for the full pod-deletion → recovery path is tracked in #172 (unit-tested at the helper level + verified by review; a full test needs a kind/fake-apiserver harness).
- Minor: suggested-port parsing is coupled to the backend error wording (fail-safe); web saved-forwards write is read-modify-write (non-atomic across concurrent tabs); no separate friendly-name field on save.

### DIFF
--- a/apps/desktop/src/components/ForwardDialog.test.tsx
+++ b/apps/desktop/src/components/ForwardDialog.test.tsx
@@ -2,12 +2,19 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 import { render, screen, fireEvent, waitFor } from "@testing-library/react";
 import React from "react";
 
-const { startPortForwardMock } = vi.hoisted(() => ({ startPortForwardMock: vi.fn() }));
+const { startPortForwardMock, saveForwardMock } = vi.hoisted(() => ({
+  startPortForwardMock: vi.fn(),
+  saveForwardMock: vi.fn(),
+}));
 vi.mock("../lib/forward", () => ({ startPortForward: startPortForwardMock }));
+vi.mock("../lib/savedForwards", () => ({ saveForward: saveForwardMock }));
 
 import { ForwardDialog } from "./ForwardDialog";
 
-beforeEach(() => startPortForwardMock.mockReset());
+beforeEach(() => {
+  startPortForwardMock.mockReset();
+  saveForwardMock.mockReset();
+});
 
 const base = { context: "kind-dev", namespace: "default", kind: "Pod", name: "web-1" };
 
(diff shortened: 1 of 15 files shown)


deveshk0:
Independent whole-branch review done. One Important finding, now fixed in `74a2d63`:

**"Give up after N attempts" was unreachable for the main failure mode.** The loop marked a session "established" on `connect_pod_api` success, but that call does no I/O (just builds an `Api<Pod>` handle) and succeeds for any valid context — so the consecutive-failure counter reset to 1 every attempt and never hit `MAX_ATTEMPTS`. A permanently-dead target (deleted Pod forward, or a Service with no recoverable pod) reconnected forever and never emitted `failed`. Fix: probe `pod_is_ready` (present + Running) before declaring the session established; a never-ready target now counts 1..MAX and gives up, while a ready-then-died target still resets and keeps recovering. Backend tests green (kube 282, streams 21).

Everything else verified sound: reconnect/re-resolve structure, listener bound once, `stop()` clean, status payloads, the pod-liveness monitor (recovery path), web `/api/settings` fetches carry cookie+CSRF, per-user settings isolation, fail-safe conflict parsing. Remaining minors are tracked (suggested-port wording coupling; non-atomic web write; no friendly-name; ForwardDialog tests use fireEvent).

Note: the full established→pod-deleted→give-up/recover path still relies on the live-cluster e2e tracked in #172 (the readiness decision reuses the unit-tested `pod_is_gone`).

Issue #147, opened 2026-07-23
enhancement(namespaces): hoist selected namespaces to the top of the selector


deveshk0:
## Summary

In the namespace selector, currently-selected namespaces stay wherever they fall in the alphabetical list, so on a cluster with many namespaces the user has to scroll/search to see or deselect what's active. Selected namespaces should be **hoisted to the top** of the dropdown for convenience.

> Note: multiple selection already exists — `NamespaceMultiSelect` supports toggling several namespaces plus "All namespaces". This ticket is specifically about **ordering**.

## Current behavior

`apps/desktop/src/ui/NamespaceMultiSelect.tsx:71-76` renders options in their original (unsorted) order; selected state only toggles a checkmark, it does not reorder:

```tsx
{namespaces.map((ns) => (
  <CommandItem key={ns} value={ns} onSelect={() => toggle(ns)}>
    <Check className={cn("size-3.5 shrink-0", selected.has(ns) ? "opacity-100" : "opacity-0")} />
    <span className="truncate">{ns}</span>
  </CommandItem>
))}
```

With dozens of namespaces, the selected ones are scattered and easy to lose track of.

## Expected behavior

- Selected namespaces appear **first**, at the top of the list (optionally under an "All namespaces" row and a subtle separator), followed by the rest.
- Within each group, keep a stable order (e.g. alphabetical, or selection order for the selected group).
- Ordering should not fight the search filter — when the user is typing in `CommandInput`, normal filtered results are fine; the hoist matters most for the unfiltered/open state.

## Proposed fix

- Compute a reordered list in the component: `[...selectedInOrder, ...unselectedSorted]`, or split into two `CommandGroup`s ("Selected" / "All") with a `CommandSeparator`.
- Guard the empty-selection case (nothing to hoist) and the all-selected case.

## Acceptance criteria

- [ ] With namespaces selected, they render at the top of the dropdown.
- [ ] Unselected namespaces follow in a stable order.
- [ ] "All namespaces" and search behavior still work.
- [ ] Test covers the reordering given a selection.

sivasubramanian95:
I'd like to work on this, planning to hoist selected namespaces via two CommandGroups per the proposed fix.

PR #187, opened 2026-08-04
feat(actions): "Copy as kubectl" command preview

### LINKED ISSUE #158: feat(actions): "Copy as kubectl" command preview for resources and actions
## Summary

Add **"Copy as kubectl"** — for any resource or action, show/copy the equivalent `kubectl` command. Helps users learn the CLI mapping, script what they just did in the UI, and audit exactly what an action will run before confirming.

## Current state

- Actions run through typed backend calls (`lib/actions.ts`, `lib/workloads.ts` → Tauri commands → `crates/kube`); the equivalent kubectl invocation is never surfaced.
- No affordance exists to copy a resource reference or an action as a CLI command.

## Proposed behavior

- On resource rows / detail: "Copy kubectl get/describe" for `kubectl get <kind> <name> -n <ns> -o yaml` etc.
- On actions (scale, restart, delete, cordon, drain, evict, cronjob trigger/suspend): show the equivalent command (e.g. `kubectl -n ns rollout restart deployment/name`) — ideally inside the confirmation dialog so users see exactly what will happen.
- Include the `--context` flag for the active context.

## Implementation notes

- Add a small pure mapper from `(action, kind, namespace, name, args)` → kubectl string; unit-test it thoroughly (it's easy to get flags subtly wrong).
- Surface via a copy button in `DetailActions` dialogs and a row/detail context menu.
- This is presentational only — it does not change how actions execute (still via the typed backend), it just displays the CLI equivalent.

## Acceptance criteria

- [ ] Resource detail/rows offer "copy as kubectl" (get/describe).
- [ ] Mutating actions show their kubectl equivalent (ideally in the confirm dialog).
- [ ] Commands include `--context` and `-n <namespace>` where applicable.
- [ ] The mapper is unit-tested across kinds/actions.

sivasubramanian95:
Closes #158.

Add "Copy as kubectl" — for any resource, pod, node, or mutating action, show or copy the equivalent kubectl command. Helps users learn the CLI mapping, script what they just did in the UI, and see exactly what an action will run before confirming it.

## What's in it

- **`toKubectl` mapper** (`lib/kubectlMapper.ts`) — a pure function from `(action, kind, namespace, name, context, ...)` to a kubectl command string. Covers get, describe, delete, scale, rollout restart, cordon, uncordon, drain, evict, and cronjob suspend/resume/trigger. Always includes `--context` and `-n <namespace>` where applicable.
- **"Copy get" / "Copy describe" buttons** on the three detail-header action bars: `PodActions`, `ResourceActions`, `NodeCordonAction`. Copies the equivalent `kubectl get ... -o yaml` / `kubectl describe ...` to the clipboard.
- **A kubectl-equivalent preview line inside every mutating confirm dialog**: pod delete/evict, resource restart/suspend/resume/trigger/scale/delete, node cordon/uncordon/drain. The scale preview only appears once the typed replica count is a valid non-negative integer — mirrors the existing guard on the actual submit.
- New shared `KubectlPreview` primitive in `src/ui`, since multiple feature files need the same small preview element and `docs/DEVELOPMENT.md` says reusable UI belongs there rather than duplicated per-view.

## Deviation from the issue

Implementation notes asked for a copy button in `DetailActions` dialogs **and** a row/detail context menu. This PR only has the first. No context-menu infrastructure exists anywhere in `ResourceBrowser.tsx` today — rows just navigate to the detail view on click, and multi-select goes through a separate `BulkActionBar`. Building one felt like a separate, larger piece of scope rather than something to fold into this PR. "Acceptance criteria" and "Proposed behavior" both phrase placement as `detail/rows`, which I've read as either being sufficient — but flagging in case you intended both and want the row-level affordance as a follow-up (or in this PR).

## Design note

The issue's wording is "get/describe" throughout, which I read as asking for both, not one merged affordance — so each header gets a separate "Copy get" and "Copy describe" button rather than one. Open to collapsing these if you'd rather keep the header slimmer.

Also skipped the debug / node-shell dialogs (ephemeral debug container, privileged node shell) — no one-line kubectl equivalent for what those do, and not in the issue's action list.

## Tests

35 new cases: 20 for the mapper across every action/kind combination (`kubectlMapper.test.ts`), 15 for the UI wiring across `DetailActions.test.tsx` and `NodeCordonAction.test.tsx` — every new button, every dialog preview, the guarded scale case, and the clipboard-rejection path (clipboard writes are wrapped in try/catch, matching `ResourceOverview.tsx`'s existing copy affordance, so a rejected permission doesn't produce an unhandled rejection or a false success toast).

## Verification

`tsc --noEmit`: clean. `pnpm test`: 106/106 files, 769/769 tests. `cargo test --workspace`: green (presentational only — no capability/backend changes).

Not covered: end-to-end verification in the running app against a live cluster — checked via unit tests only, same caveat as #147.

### DIFF
--- a/apps/desktop/src/components/CopyAsKubectlButton.test.tsx
+++ b/apps/desktop/src/components/CopyAsKubectlButton.test.tsx
@@ -1,0 +1,47 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { render, screen, fireEvent, waitFor } from "@testing-library/react";
+import React from "react";
+
+const { notifyMock } = vi.hoisted(() => ({
+  notifyMock: { success: vi.fn(), error: vi.fn(), info: vi.fn() },
+}));
+vi.mock("../lib/notify", () => ({ notify: notifyMock }));
+
+import { CopyAsKubectlButton } from "./CopyAsKubectlButton";
+
+beforeEach(() => {
+  notifyMock.success.mockReset();
+  Object.defineProperty(navigator, "clipboard", {
+    value: { writeText: vi.fn().mockResolvedValue(undefined) },
+    configurable: true,
+  });
+});
+
+describe("CopyAsKubectlButton", () => {
+  it("copies the get command (as yaml) from the menu", async () => {
+    render(<CopyAsKubectlButton kind="Pod" name="web-1" namespace="default" context="kind-dev" />);
+    fireEvent.click(screen.getByRole("button", { name: "Copy as kubectl" }));
+    fireEvent.click(await screen.findByRole("button", { name: "Copy get" }));
+    expect(navigator.clipboard.writeText).toHaveBeenCalledWith(
+      "kubectl get pods web-1 -n default --context kind-dev -o yaml",
+    );
+    await waitFor(() => expect(notifyMock.success).toHaveBeenCalledWith("Copied kubectl command"));
+  });
+
+  it("copies the describe command from the menu", async () => {
+    render(<CopyAsKubectlButton kind="Pod" name="web-1" namespace="default" context="kind-dev" />);
+    fireEvent.click(screen.getByRole("button", { name: "Copy as kubectl" }));
+    fireEvent.click(await screen.findByRole("button", { name: "Copy describe" }));
+    expect(navigator.clipboard.writeText).toHaveBeenCalledWith(
+      "kubectl describe pods web-1 -n default --context kind-dev",
+    );
+    await waitFor(() => expect(notifyMock.success).toHaveBeenCalledWith("Copied kubectl command"));
+  });
+
+  it("closes the menu after copying", async () => {
+    render(<CopyAsKubectlButton kind="Node" name="node-1" context="kind-dev" />);
+    fireEvent.click(screen.getByRole("button", { name: "Copy as kubectl" }));
+    fireEvent.click(await screen.findByRole("button", { name: "Copy get" }));
+    await waitFor(() => expect(screen.queryByRole("button", { name: "Copy get" })).toBeNull());
+  });
+});
--- a/apps/desktop/src/components/CopyAsKubectlButton.tsx
+++ b/apps/desktop/src/components/CopyAsKubectlButton.tsx
@@ -1,0 +1,58 @@
+import React, { useState } from "react";
+import { Copy, ScrollText } from "lucide-react";
+import { Popover, PopoverContent, PopoverTrigger } from "@/components/ui/popover";
+import { Button } from "@/components/ui/button";
+import { toKubectl } from "../lib/kubectlMapper";
+import { copyKubectlCommand } from "../lib/copyKubectl";
+
+const MENU_ITEM_CLASS =
+  "flex w-full items-center gap-2 rounded px-2 py-1.5 text-left text-sm hover:bg-accent";
+
+/**
+ * Header affordance for copying a resource's kubectl get/describe command.
+ * Collapses what used to be two separate icon buttons into one, opening a
+ * small menu — headers were getting crowded, and Copy/ScrollText don't
+ * self-describe as get-vs-describe on their own.
+ */
+export function CopyAsKubectlButton({
+  kind,
+  name,
+  namespace,
+  context,
+}: {
+  kind: string;
+  name: string;
+  namespace?: string | null;
+  context: string;
+}) {
+  const [open, setOpen] = useState(false);
+
+  async function copy(action: "get" | "describe") {
+    await copyKubectlCommand(
+      toKubectl({ action, kind, name, namespace, context, output: action === "get" ? "yaml" : undefined }),
+    );
+    setOpen(false);
+  }
+
+  return (
+    <Popover open={open} onOpenChange={setOpen}>
+      <PopoverTrigger asChild>
+        <Button type="button" variant="ghost" size="icon-sm" aria-label="Copy as kubectl" title="Copy as kubectl">
+          <Copy aria-hidden="true" />
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent align="start" className="w-44 p-1">
+        <div role="group" aria-label="Copy as kubectl">
+          <button type="button" className={MENU_ITEM_CLASS} onClick={() => void copy("get")}>
+            <Copy className="h-4 w-4" aria-hidden="true" />
+            Copy get
+          </button>
+          <button type="button" className={MENU_ITEM_CLASS} onClick={() => void copy("describe")}>
+            <ScrollText className="h-4 w-4" aria-hidden="true" />
+            Copy describe
+          </button>
+        </div>
+      </PopoverContent>
+    </Popover>
+  );
+}
--- a/apps/desktop/src/components/DetailActions.test.tsx
+++ b/apps/desktop/src/components/DetailActions.test.tsx
@@ -149,30 +149,39 @@ describe("PodActions", () => {
     await waitFor(() => expect(onDeleted).toHaveBeenCalled());
   });
 
-  it("copies the kubectl get and describe commands for the pod", () => {
+  it("copies the kubectl get command for the pod via the Copy as kubectl menu", async () => {
     render(<PodActions context="kind-dev" pod={pod} onDeleted={() => {}} />);
-    fireEvent.click(screen.getByRole("button", { name: "Copy get" }));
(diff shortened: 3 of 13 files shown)


deveshk0 (changes requested):
Nice piece of work — the mapper as a pure, separately-tested function is the right shape, and a few details show real care: `validReplicas` mirroring `doScale`'s own guard instead of previewing `--replicas=NaN`, the clipboard `try/catch` that avoids a false success toast, and `rollout-restart` mapping exactly onto what the backend patches (`kubectl.kubernetes.io/restartedAt`). The `-n` omission for cluster-scoped kinds is also correct — it works because `ResourceBrowser.tsx:837` passes `namespace: null` for them, which is easy to get wrong.

My findings cluster in one place: the previews render **inside destructive-action confirm dialogs**, so a string that doesn't match what the app actually does is misleading at the worst possible moment. Two of these I'd want fixed before merge; the rest are cheap or deferrable.

A note on the test suite: the mapper tests assert literal output strings, so they pin whatever the mapper emits — right or wrong. I checked each mapping against the backend capability it claims to mirror rather than relying on green tests. (I did run them: 67/67 pass across the three files.)

---

### 🔴 1. `evict` maps to a force-delete, contradicting the dialog it sits under

`toKubectl` renders `evict` as:

```
kubectl delete pod <name> --grace-period=0
```

The backend does something quite different — `k8s.evictPod` calls the **Eviction API**, and its own summary is *"evict a pod via the eviction API (respects PodDisruptionBudgets)"* (`crates/kube/src/actions.rs:70,77`).

The preview appears directly beneath dialog text reading *"Gracefully evict `<pod>` … (respects disruption budgets)"* — and shows a command that does neither:

- `delete` bypasses PodDisruptionBudgets entirely.
- kubectl's own help documents `--grace-period` as *"Can only be set to 0 when `--force` is true"*, so depending on version you get a rejection or a silently-clamped grace period. Never an eviction.

There genuinely is no one-liner kubectl equivalent for a single-pod eviction, so there's no clean fix — the honest options are to omit the preview for `evict`, show the `--raw` eviction-subresource call, or at minimum drop `--grace-period=0` so it's a plain graceful delete and label it as an approximation. Happy with any of those; the current string is the one option I'd rule out.

`kubectlMapper.test.ts:100-104` pins the current behaviour, so that test wants updating alongside.

### 🔴 2. The Node detail header renders "Copy get" / "Copy describe" twice

`ResourceBrowser.tsx:925-939` renders `NodeCordonAction` **and** `ResourceActions` for `kind === "Node"`, and this PR adds the same pair of buttons to both. Rendering the two together the way `ResourceBrowser` does gives:

```
Copy get | Copy describe | Cordon | Drain | Copy get | Copy describe | Delete
```

<details>
<summary>How I checked</summary>

Rendered `<NodeCordonAction context="prod" name="node-1" />` and `<ResourceActions context="prod" kind="Node" namespace={null} name="node-1" />` as siblings (mirroring `ResourceBrowser.tsx:925-939`), waited for the node object to resolve so `NodeCordonAction` passes its `cordoned === null` guard, then enumerated buttons by `aria-label`.

</details>

Both components' own tests pass because each renders in isolation — exactly the class of bug component-scoped tests can't see. Simplest fix: drop the two buttons from `NodeCordonAction`, since `ResourceActions` already covers Node.

### 🟠 3. `cronjob-trigger` previews a Job name the app never creates

Preview: `kubectl create job --from=cronjob/<name> <name>-manual`

The app sends a `Date.now()` suffix (`lib/actions.ts:105`) and the backend builds `<cronjob>-<suffix>` (`crates/kube/src/cronjobs.rs:31`), plus a non-controller `ownerReference` that `kubectl create job --from` doesn't set. So the previewed command differs from what runs, and copying it works exactly once — a second run fails with `AlreadyExists`, where the app's timestamped name never collides. `<name>-manual-$(date +%s)` would close most of the gap.

### 🟡 4. `drain` overstates on one axis, understates on another

`--ignore-daemonsets --delete-emptydir-data` is a well-judged approximation of the backend's skip-DaemonSet/mirror-pod loop (`crates/kube/src/actions.rs:369-397`) — that one isn't obvious and you got it right. But the backend also evicts *unmanaged bare pods*, which `kubectl drain` refuses without `--force`. So a copied preview can fail where the app succeeded. Worth a `--force`, or a note that it's approximate.

### 🟡 5. No shell quoting

`name` / `namespace` / `context` are interpolated raw into a string whose entire purpose is being pasted into a shell. Resource names are DNS-safe, but kubeconfig **context names are arbitrary** and can contain spaces — `--context my cluster` silently becomes two arguments. Quoting anything outside `[A-Za-z0-9._@:/-]+` would cover it.

### 🟡 6. The cronjob patch previews shell-specific quoting

`-p '{"spec":{"suspend":true}}'` is fine in bash/zsh/PowerShell, broken in `cmd.exe` — and we ship Windows builds.

---

### Smaller things

- **`kindLower` reimplements a table we already have.** `DetailActions.tsx` already imports `kindToResource` from `lib/access` (`access.ts:167-191`), which holds the authoritative kind→resource mapping. `kind.toLowerCase()` works for today's kinds but will drift for anything where the resource name isn't just the lowercased kind.
- **Three copies of `copyKubectl`.** Byte-identical `try/catch` + `notify.success` in `PodActions`, `ResourceActions`, and `NodeCordonAction` — one exported helper beside `toKubectl` would do.
- **The preview is unlabelled and uncopyable.** `KubectlPreview` is a bare `<code>` in muted xs text with no "Equivalent kubectl:" prefix, so inside a dialog it reads as part of the message. And the *mutating* previews can't be copied by clicking — only get/describe got buttons. Given #158's "script what they just did in the UI", those are arguably the ones people most want to copy.

---

### On your two open questions

**Row/detail context menu deferred** — agreed, and I'd keep it out. There's no context-menu infrastructure in `ResourceBrowser.tsx` today; building one would dwarf this change and bury it. Let's spin it out as a follow-up issue.

**Separate "Copy get" / "Copy describe" buttons** — I'd collapse them. `PodActions` is up to nine header buttons now, and `Copy` / `ScrollText` don't self-describe as get-vs-describe. One "Copy as kubectl" affordance (a small menu, or just `get -o yaml`) keeps the header readable.

Skipping the debug / node-shell dialogs was the right call.

---

**Requesting changes on 1 and 2** — both are small, and both make the feature actively wrong rather than merely incomplete. Worth folding in 3 while you're there. Everything from 4 down is fine as follow-up if you'd rather land this.

sivasubramanian95:
Thanks for the thorough review — fixed everything, including the items you'd flagged as deferrable. Point by point:

**1. `evict` command** — Went with dropping the preview and showing an explanatory note instead of a command (`No single-line kubectl equivalent — eviction uses the pod's /eviction subresource, which respects PodDisruptionBudgets`), rather than the `--raw` call or the labeled-approximation delete. Figured a wrong-looking multi-line raw API call would be worse for a UI that's supposed to keep things copy-pasteable, and a labeled approximation still risks someone copying a command that skips PDBs.

**2. Duplicate Node buttons** — Dropped the Copy get/describe buttons from `NodeCordonAction`; `ResourceActions` already covers Node.

**3. cronjob-trigger name** — Now previews `<name>-manual-$(date +%s)`, matching your suggestion.

Folded in 4–6 and the smaller items too:
- `drain` now includes `--force`, matching the backend's eviction of unmanaged bare pods.
- `namespace`/`context`/`name` are shell-quoted (double-quoted, not single — cmd.exe doesn't treat `'` as a quote character, so single-quoting would've reintroduced the same problem as #6 for context names with spaces).
- cronjob patch body switched to double-quoted/escaped JSON, valid across bash/zsh/PowerShell/cmd.exe.
- `kindLower` now comes from `kindToResource` instead of a second hand-rolled table.
- `copyKubectl` deduped into one `lib/copyKubectl.ts` helper.
- `KubectlPreview` now labels the command ("Equivalent kubectl:") and every dialog preview is copyable, not just get/describe.

On the two open questions: kept the context menu deferred, and collapsed Copy get/describe into a single "Copy as kubectl" menu (`CopyAsKubectlButton`) per your suggestion, rather than defaulting to `get -o yaml` — figured it also gives us one shared place to make the delete/evict/etc. dialog previews copyable, which was on your smaller-things list anyway.

784/784 tests pass locally, coverage's still above the floor.

deveshk0:
@sivasubramanian95, don't worry about comments; I am just trying some automated reviews.

deveshk0 (approved):

PR #204, opened 2026-08-11
refactor(mcp): rename kind resolver API

### LINKED ISSUE #199: MCP: rename McpServer::resources to kind_resolver
Follow-up from #194 (MCP resources, #24). Naming only, no behaviour change.

`McpServer` has a field `resources`, a builder `with_resources(...)` and a getter `resources()` — all of which hold and return an `Arc<dyn KindResolver>`, not resources:

```rust
pub fn with_resources(mut self, kinds: Arc<dyn resources::KindResolver>) -> Self
pub fn resources(&self) -> &Arc<dyn resources::KindResolver>
```

The name came from the module (`crate::resources`) rather than from what the value is. It reads as "the server's resources", which is misleading in a codebase that now has actual MCP resources, resource URIs, resource templates and resource reads. The next person adding a genuine resource provider will reach for this name and find a kind resolver.

Rename to `with_kind_resolver` / `kind_resolver()` with a matching field name. Call sites are few:

- `crates/mcp/src/lib.rs` — field, builder, getter, and the fail-closed `NoKinds` default in `new`
- `crates/mcp/src/stdio.rs` — `server.resources().as_ref()` in the `resources/read` arm and in `handle_subscription`
- `apps/desktop/src-tauri/src/mcp.rs` and `apps/desktop/src-tauri/src/main.rs` — three `.with_resources(srelens_registry::kind_resolver())` chains, which currently read oddly precisely because the argument already carries the better name
- test helpers in `crates/mcp/src/stdio.rs`

Note `with_watcher` / `watcher()` alongside it are named correctly and need no change.

sidsri14:
Closes #199.

## Summary

- rename `McpServer`'s injected `resources` field to `kind_resolver`
- rename the public builder/getter to `with_kind_resolver` and `kind_resolver`
- update MCP, desktop, and e2e call sites without changing resource behavior
- add an API-level regression test for the renamed builder and getter

The old name described the module rather than the value it stored. The new API makes it explicit that this dependency resolves Kubernetes kind scope and leaves the actual MCP resource and watcher APIs distinct.

## Validation

- `cargo test -p srelens-mcp` - 216 passed
- `cargo check -p srelens-desktop --tests` - passed
- `pnpm test` - 127 files and 992 tests passed; 84.88% line coverage
- `pnpm build` - passed
- `git diff --check origin/dev...HEAD` - passed

`cargo test --workspace` also completed every affected MCP and desktop suite. Eight `srelens-kube` tests fail only on Windows because they assert POSIX separators or `/bin/helm`; the identical eight failures were reproduced on untouched current `origin/dev` (283 passed, 8 failed in that package).

## AI assistance

Implemented with OpenAI Codex assistance. I reviewed the conflict resolution, final diff, and verification results.


deveshk0:
Hi @sidsri14, could you please resolve the conflicts and mark it ready?

We would put this in v0.6

sidsri14:
Resolved the conflict against current dev, updated the new shared server-construction call sites, reran the affected Rust and full frontend gates, and marked this ready for review. No force push was used.

deveshk0 (approved):

Issue #198, opened 2026-08-08
MCP: audit-log resources/subscribe and unsubscribe


deveshk0:
Follow-up from #194 (MCP resources, #24). Raised by the final whole-branch review.

Every capability-backed resource read writes an audit record naming the underlying capability, with `is_sensitive`-driven redaction — verified in review. `resources/subscribe` writes **nothing**.

That is a real observability gap rather than a security hole. A subscribe opens a live watch against the cluster and can be repeated up to the 32-subscription cap, so an agent can establish and hold 32 concurrent watches with no trace in the audit log. An operator reading the log sees the reads an agent performed but not that it is also holding open watches, or on which objects. `resources/unsubscribe` is likewise unrecorded, so the log cannot show when a watch was released.

No content flows over this path — notifications carry only the URI and the client re-reads through the audited read path — so this is metadata only. That is why it was filed rather than fixed in #194.

Implementation notes:

- `handle_subscription` in `crates/mcp/src/stdio.rs` is synchronous, and `AuditSink::record` is synchronous too, so this needs no restructuring.
- The transport is always `Transport::Stdio` here (subscribe is stdio-only by design; HTTP advertises `subscribe: false`).
- `AuditRecord.tool` is typed for a capability id. A subscribe invokes no capability, so decide deliberately what goes there — the method name (`resources/subscribe`) is the honest answer, but it means the field no longer always holds a capability id, and anything parsing the log should keep working. Worth checking `crates/mcp/src/audit.rs` and the log-tailing UI before choosing.
- Record the outcome, not just the attempt: a subscribe rejected for the cap, for a non-subscribable URI, or by a watcher error is as interesting as a successful one.
- The URI belongs in the record. It is not sensitive (Secrets are not addressable at all), but it should go through `redact` like any other args value for consistency.

deveshk0:
Implemented by #283 and released in **v0.7.2**. The issue stayed open only because that PR said "Implements #198" — not one of GitHub's closing keywords — so nothing auto-closed it.

Verified against the acceptance notes above:

- **9 audit records** across `handle_subscription`, covering both methods.
- **Outcomes, not just attempts** — 7 error paths (bad URI, cap reached, non-subscribable kind, watcher failure) and 2 success paths.
- **`tool` carries the method name** (`resources/subscribe` / `resources/unsubscribe`), the honest answer the notes settled on.
- **The URI goes through `redact`** in all 9, for consistency with every other record.

[...]

PR #283, opened 2026-08-18
fix(mcp): audit-log resources/subscribe and unsubscribe

sidsri14:
Implements #198.

## What

`resources/subscribe` and `resources/unsubscribe` were the last
capability-backed MCP paths that wrote no audit record — an agent could
hold up to `MAX_SUBSCRIPTIONS` concurrent watches with nothing left in
the log. This records every subscribe/unsubscribe outcome in
`handle_subscription` (`crates/mcp/src/stdio.rs`), matching the existing
`resources/read` convention:

- `transport` is always `Stdio` (the HTTP transport advertises
  `subscribe: false`).
- `decision` is always `"auto"` (the path is not consent-gated).
- `tool` holds the method name (`resources/subscribe` /
  `resources/unsubscribe`).
- The URI goes through `redact` for consistency, even though Secrets are
  not addressable so nothing here is sensitive.
- The outcome is recorded on **every** branch, so a rejected subscribe
  (non-subscribable URI, unaddressable Secret, watcher refusal, the
  32-watch cap, a watch that dies immediately) shows up as loudly as a
  successful one.

## Tests

Six new tests in `crates/mcp/src/stdio.rs`, written first (TDD): a
successful subscribe, a rejected subscribe, an unsubscribe, plus the
rejection branches — non-subscribable URI, unaddressable Secret, watcher
error, and the watch cap. All pass; full crate suite is 234/234.

## Verification / CI

CI gates the backend on `cargo llvm-cov --fail-under-lines 85`; every new
line is exercised by the new tests, so coverage can only hold or rise.

Note: local verification ran on the `x86_64-pc-windows-gnu` toolchain
(MinGW/mingw-w64) because the MSVC toolchain lacks the Windows SDK here.
Repo CI runs on Linux stable and is authoritative.

### DIFF
--- a/crates/mcp/src/stdio.rs
+++ b/crates/mcp/src/stdio.rs
@@ -411,17 +411,43 @@ pub(crate) fn handle_subscription(
 
     let uri = match crate::resources::ResourceUri::parse(uri_str) {
         Ok(u) => u,
-        Err(message) => return Some(err(id, -32602, &message)),
+        Err(message) => {
+            server.audit().record(crate::audit::AuditRecord {
+                transport: crate::Transport::Stdio,
+                tool: method.to_string(),
+                args: crate::audit::redact(&json!({"uri": uri_str}), false),
+                decision: "auto",
+                outcome: "error",
+                error: Some(message.clone()),
+            });
+            return Some(err(id, -32602, &message));
+        }
     };
     // Canonical form, so two spellings of one object cannot make two watches.
     let canonical = uri.to_string();
 
     if method == "resources/unsubscribe" {
         subs.remove(&canonical);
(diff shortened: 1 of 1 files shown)


deveshk0 (approved):
Thanks for this — closing the last unaudited path, and the tests are the best part.

Approving. Every exit records, including both `watch ended immediately` branches, which are the easy ones to miss. `Transport::Stdio` and `decision: "auto"` both check out.

Non-blocking: the record block repeats eight times, differing only in `outcome`/`error`. A closure over the invariant parts would stop a ninth branch forgetting to record — the exact bug this PR fixes. Not needed here.

Scope note: I read the implementation in full and the test names, not all six test bodies.

PR #302, opened 2026-08-19
fix(mcp): attribute subscription audit records to the calling transport

deveshk0:
Every `resources/subscribe` and `resources/unsubscribe` made by a networked MCP client was audit-logged as if it had arrived over stdio.

## The bug

`handle_subscription` (`stdio.rs`) stamped all nine audit records it emits with a hardcoded `Transport::Stdio`. But both transports route through that helper — `http.rs:352` dispatches the two subscription methods there deliberately, because they need the caller's own push machinery (session registry, dirty set, stream wakeup) rather than the generic `handle_request` path.

Attribution is the entire purpose of the audit log, which makes a wrong transport worse than a missing record. It reads as authoritative, sends an operator filtering by transport to the wrong caller, and silently omits exactly the remote traffic the log exists to account for. On a change whose own issue (#198) is labelled `type/security`, that matters.

## The fix

The caller's transport becomes a parameter: `Transport::Http` from the HTTP route, `Transport::Stdio` from the stdio serve loop, used by all nine records. No behaviour changes beyond what the records say.

## Verification

- New test `a_subscription_over_http_is_audited_as_http` opens an SSE stream, subscribes over HTTP, and asserts every resulting record names `Http`. It fails on the previous code with `left: Stdio, right: Http`.
- `cargo test -p srelens-mcp` — 239 passed.
- `cargo test --workspace` — **1091 passed, 0 failed**; `cargo check --workspace --all-targets` clean, no unused warnings.

## How it reached dev

Raised by automated review on #283 and marked resolved there without the change being made, so it merged. Worth knowing that thread resolution on that PR did not imply the finding was addressed.

Given #283 is already on `dev` and unreleased, merging this before the next promotion keeps the defect out of a shipped version entirely.

### DIFF
--- a/crates/mcp/src/http.rs
+++ b/crates/mcp/src/http.rs
@@ -359,7 +359,12 @@ async fn rpc(
         // never deliver.
         let sessions = st.push.sessions.lock().unwrap_or_else(|e| e.into_inner());
         match sessions.get(&session) {
-            Some(c) => handle_subscription(&st.server, &c.subs, &c.dirty, &c.wake, &req, method),
+            Some(c) => {
+                handle_subscription(
+                    &st.server, &c.subs, &c.dirty, &c.wake, &req, method,
+                    crate::Transport::Http,
+                )
+            }
             None => req.get("id").cloned().map(|id| {
                 json!({"jsonrpc": "2.0", "id": id, "error": {"code": -32002, "message":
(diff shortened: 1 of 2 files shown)


Parul1411 (approved):
LGTM

Issue #305, opened 2026-08-20
Epic: migrate srelens to the new design


deveshk0:
Tracking issue for moving srelens onto the new visual and interaction language prototyped in `srelens-new-design/` (10,758 lines of React on dummy data).

## Decisions

| Decision | Choice |
|---|---|
| End state | The old design is deleted. The switch is temporary scaffolding. |
| Who can switch | Users, via a shipped Settings toggle, from the first release. |
| Where changes land | A ported screen is developed only in the new design; the old one is frozen. |
| Code structure | Shared service package + shared component kit; both designs consume them. |

## The constraint that shapes everything

**The two stylesheets cannot share a document.** The old app loads `tailwindcss` + `tw-animate-css` + `shadcn/tailwind.css` plus 4,539 lines of `fl-*` rules; the new design loads `tailwindcss` plus 1,053 lines of its own, uses `[data-theme=dark]` where the old uses `.dark`, and both write global `body`/`*` rules.

So **old and new screens cannot render in the same window.** A per-screen migration inside one shell is not available. The switch is at the root, one stylesheet per document, and "gradual" means porting screens into the new tree until it reaches parity — not mixing them live.

## Two component libraries, merged per component

`apps/desktop/src/ui` has 22 components, 2,112 lines, 17 test files — mature and behaviour-rich. The mock has 64 components across 19 modules — broader, and the only source for the new design's novel surfaces (ConsoleDock, DrillCard, Inspector, ResourceTree, TabStrip, Titlebar, StatusBar, ContextMenu, ClusterRail), but with no tests, states or keyboard handling.

- **Both have it** → keep the old API, behaviour and tests; take the mock's visuals.
- **Only the mock has it** → take its component, then harden it.
- **Only the old kit has it** → restyle against the new tokens; API and tests unchanged.

## Steps

- [x] 1. Extract `@srelens/core` — the service layer *(shipped in #310, #311)*
  - [x] #306 fix the two dependencies that point the wrong way
  - [x] #307 create the package, move `transport/`
  - [x] #308 move `lib/` in, behind two entry points
  - [ ] #309 record the package boundary in a README (CI half shipped in #311)
- [x] 2. `packages/ui-kit` + `packages/ui-next` scaffold, and the switch
  - [x] #312 the switch: preference, dynamic stylesheet + tree, placeholder, Settings toggle *(shipped in #314)*
  - [x] the component-library merge into `packages/ui-kit`
[...]

PR #310, opened 2026-08-20
refactor(core): fix the two dependencies that point the wrong way

### LINKED ISSUE #306: refactor(core): fix the two dependencies that point the wrong way
Part of #305, step 1. Prerequisite for extracting the service layer — and worth landing on its own regardless.

Two modules in `lib/` depend on things above them, which blocks `lib/` from becoming a package.

## 1. `lib/paletteActions.ts` imports a UI component to read a lookup table

`ResourceKind`, `RESOURCE_LABELS` and `K8S_KIND` live in `components/ResourceBrowser.tsx` (lines 86-129, 131-175, 177-221). They are pure data. Eleven files import them.

Move them to `lib/kinds.ts` and repoint all eleven. Move the declarations verbatim — the union has 40+ members and a retyped one becomes a silent runtime hole.

## 2. `lib/openTabs.ts` imports the app root for a type

`ViewTab` is declared in `App.tsx:94-114`, so the module that *persists* a tab imports the app root to describe its own payload. Three files use it.

Move it to `lib/tabs.ts`.

## Approach

Land these as three commits so each is reviewable on its own:

1. Create `lib/kinds.ts`, have `ResourceBrowser` re-export from it, repoint `paletteActions`. Keeps the blast radius to one file.
2. Repoint the other ten importers; drop the temporary re-export.
3. Create `lib/tabs.ts`, repoint `openTabs`, drop the compatibility re-export.

## Done when

```bash
# lib depends on nothing above it
cd apps/desktop && grep -rn 'from "\.\./App"\|from "\.\./components\|from "\.\./ui' \
  src/lib/*.ts | grep -v '\.test\.'     # no output

npx tsc --noEmit && npx vitest run       # 1248 passed (1245 + 3 new), 0 failed
```

New tests: `lib/kinds.test.ts` asserting `K8S_KIND` maps the common kinds, that `overview` maps to `""` because it is a view not a resource, and that every kind in `K8S_KIND` has a label in `RESOURCE_LABELS` — the two tables must stay in step or a kind renders blank in the sidebar.

## Not in scope

No behaviour changes. This is a move plus an import rewrite; a red test means the move was wrong, not that the test needs updating.

deveshk0:
Closes #306. Part of #305, step 1 — the prerequisite for extracting the service layer, and worth landing on its own regardless.

Two modules in `lib/` depended on things above them, which is what blocks `lib/` from becoming a package.

## `lib/paletteActions.ts` imported a UI component to read a lookup table

`ResourceKind`, `RESOURCE_LABELS` and `K8S_KIND` lived in `components/ResourceBrowser.tsx`. They are pure data — a kind → Kubernetes kind map and a kind → label map — and eleven files used them, so a palette module reached into a React component to read a table.

They now live in `lib/kinds.ts`, moved verbatim rather than retyped: the union has 40+ members and a retyped one becomes a silent runtime hole.

## `lib/openTabs.ts` imported the app root for a type

`ViewTab` was declared in `App.tsx`, so the module that *persists* a tab imported the app root to describe its own payload. It now lives in `lib/tabs.ts`.

## Landed as three commits

1. `lib/kinds.ts` created; `ResourceBrowser` re-exports from it; `paletteActions` repointed — blast radius of one file.
2. The other importers repointed; the temporary re-export dropped.
3. `lib/tabs.ts` created; `openTabs` repointed.

## Worth a reviewer's eye

`App.test.tsx` mocked `ResourceBrowser` to supply a **reduced set of five kinds**. Once `App.tsx` read the tables from `lib/kinds`, that mock stopped applying and the real forty leaked in. The mock moves to `lib/kinds` alongside the tables — otherwise the sidebar assertions would have silently started testing something other than what they were written for. This is the only non-mechanical part of the change.

## Verification

```
grep -rn 'from "../App"|from "../components"|from "../ui"' src/lib/*.ts   # no output
npx tsc --noEmit                                                          # clean
npx vitest run --coverage                    # 1248 passed, 85.89% lines
```

**1248 = 1245 + 3 new**, matching the plan's prediction. New tests in `lib/kinds.test.ts` assert the common kind mappings, that `overview` maps to `""` because it is a view not a resource, and that every kind in `K8S_KIND` has a label in `RESOURCE_LABELS` — the two tables must stay in step or a kind renders blank in the sidebar.

No behaviour changes: this is a move plus an import rewrite, and the suite is the proof.


navneet066 (approved):
LGTM

PR #354, opened 2026-08-26
feat(ui-next): the helm screen

deveshk0:
Ports Helm onto `packages/ui-next` — the last of step 7's three specs, and the last of classic's bottom-dock kinds. Closes the dock: `logs` went in #344, `terminal`/`shell` in #352, `helm` here.

This is the only genuinely destructive surface in the migration, so most of the work below is about not lying to the reader about what is going to happen.

## What ships

- **A status verdict in core** (`helmStatus`) — helm's set is a passthrough string, not a closed enum, so an unrecognised status renders as the word helm gave, toned neutral. Not healthy, not broken.
- **A `revision` parameter** through Rust and core, so a release can be read at an earlier revision. A missing revision errors; there is deliberately no fallback, because falling back to the current one would diff a manifest against itself and report "no changes" about a release that changed.
- **`DiffLines` in the kit** — a presentational diff over `DiffRow[]`. Classic's `DiffView` did **not** move: it takes a `DiffDoc` and renders a manifest dry-run, which is a different component for a different job.
- **A module-level operation store** — `helm upgrade --wait` runs for minutes and outlives the dialog that started it. This is the third such store, after port-forwards and terminal sessions.
- **The four-mode dialog** (install / upgrade / rollback / uninstall) with two gates classic never had: uninstall requires the release name typed exactly, rollback asks once.
- **The operation-or-diff pane**, a failures register, the screen, two status-strip segments, and routing.

## Things worth a reviewer's attention

**Nothing offers to cancel, and the reason is not the obvious one.** The store's own comment used to say that closing a handle "touches the release not at all". That was false: closing aborts the task owning helm's child process, spawned `kill_on_drop(true)`, so it SIGKILLs helm wherever it had reached. No live path reaches it today; the comment now states it as a rule to keep, and a test holds the export list so a new export cannot appear without its author reading that rule.

**A failed operation stays until dismissed, and is readable wherever the strip counts it.** The strip's principle is that a segment counts what the screen it opens can show — so `/helm` grew a failures register scoped exactly as the segment is, and by nothing else.

**Copy that claimed more than srelens knows.** Four instances found and fixed: a rollback hint saying "no history" when history existed; the uninstall alert stating object counts never fetched; a footer naming a revision the dialog could not honour; and an `Equivalent command` omitting the `--values` the backend appends — copy that line into a terminal and you got an upgrade with no values.

**`Upgrade` used to reset a release to chart defaults.** It sent no values, and helm's default for that is to discard the user's. Classic seeded the editor from the release; this now does too.

## Also fixed here, pre-existing and separable

- **"No cluster in focus" shown when a cluster was in focus** — the contexts store had no loaded-or-failed signal, so "still loading" and "the listing failed" both rendered as an accusation. A rejecting `listContexts` was swallowed entirely. Affects every screen, not just this one.
- **Tables wobbled and scrolled unevenly** — `.tbl td` had no tabular figures, so one changed digit in a live CPU or memory cell dragged every digit before it (measured: 2.85px). Separately the virtualizer sized the list from the first rendered row, which sits under a spacer and is 0.375px short — swinging the scroll extent by 545px on 1500 rows. Both arrived with the Table port, not with this branch.

## Verification

`vitest` 4659 passing; coverage 92.18 / 88.35 / 81.65 against floors 85 / 80 / 76; `pnpm -r typecheck`; 568 cargo tests; `cargo build`; `pnpm build`.

Every task ran a mutation pass. Those caught three real bugs rather than only weak tests: a "no changes" state that printed the whole manifest back as context, a re-list that skipped failed operations, and a dismiss that cleared the register while leaving the strip counting.

The status strip was measured at the 960px minimum window rather than estimated — it overflowed at 1036px and clipped `Ask` unreachably, since `body` is `overflow: hidden`. Fixed with the tab strip's own scroll policy.

## Not photographed

The demo cluster has zero Helm releases and nothing was installed to make a screenshot prettier. The populated table and the clipped action column were caught from a real cluster with 383 releases; the diff pane at 11px remains unverified by eye.

## Follow-ups filed, not smuggled in

Dialogs are window-modal and block tab switching — its own PR next, and note it invalidates a wrong-cluster clearance this review relied on. Plus: `sessions.ts` has the handle-after-settle leak this branch fixed in its own copy; the kit's `Field` folds its hint into the label's accessible name; `tokens-only.test.ts` catches hex but not `rgb()` or palette classes; the table's resize observer has never attached, so auto-sized tables never re-fit.

### DIFF
--- a/apps/desktop/src/components/AppearanceSettingsSection.test.tsx
+++ b/apps/desktop/src/components/AppearanceSettingsSection.test.tsx
@@ -91,6 +91,15 @@ describe("AppearanceSettingsSection", () => {
     expect(items).toContain("Terminals");
   });
 
+  it("lists Helm, whose operations run on after the dialog that started them", () => {
+    // The release table, the diff pane and the four operations. An upgrade or
+    // a rollback outlives its dialog, so a reader weighing the toggle is
+    // weighing whether they can drive Helm at all in the new design.
+    render(<AppearanceSettingsSection />);
+    const items = screen.getAllByRole("listitem").map((li) => li.textContent);
+    expect(items).toContain("Helm");
+  });
+
   it("lists the toolbox, the only screen that is about the machine and not a cluster", () => {
     // The managed kubectl, helm and krew under ~/.srelens/bin, plus what a
     // context's exec-auth needs. Nothing about it is cluster-scoped, so a
--- a/apps/desktop/src/design.test.ts
+++ b/apps/desktop/src/design.test.ts
@@ -64,6 +64,7 @@ describe("the list of ported screens", () => {
       "/logs",
       "/forwards",
       "/terminals",
+      "/helm",
       "/toolbox",
     ]);
   });
@@ -79,6 +80,7 @@ describe("the list of ported screens", () => {
       "Logs",
       "Port forwards",
       "Terminals",
+      "Helm",
       "Toolbox",
     ]);
   });
--- a/apps/desktop/src/design.ts
+++ b/apps/desktop/src/design.ts
@@ -224,6 +224,11 @@ export const PORTED_SCREENS: ReadonlyArray<{ route: string; name: string }> = [
   // because it shares their shape — a session outlives the tab that started
   // it, so this screen is where the ones already running are seen and ended.
   { route: "/terminals", name: "Terminals" },
+  // The release table, the rendered diff between two revisions, and the four
+  // operations. Listed with forwards and terminals because it shares their
+  // shape: an upgrade or a rollback outlives the dialog that started it, and
+  // the status strip counts the ones still running.
+  { route: "/helm", name: "Helm" },
   // The only screen here that is about the machine rather than a cluster: the
   // managed kubectl, helm and krew under ~/.srelens/bin, and what the active
   // context's exec-auth needs on PATH.
--- a/crates/kube/src/helm.rs
+++ b/crates/kube/src/helm.rs
@@ -149,6 +149,11 @@ pub struct GetHelmReleaseIn {
     pub context: String,
     pub namespace: String,
     pub name: String,
+    /// The revision to read. Omitted means the current one, exactly as
+    /// before this field existed — every caller that doesn't know about
+    /// revisions keeps getting what it always got.
+    #[serde(default)]
+    pub revision: Option<i64>,
 }
 
 #[derive(Debug, Clone, Serialize, JsonSchema)]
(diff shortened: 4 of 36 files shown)


deveshk0:
Went through the CodeQL review. **None of the six alerts are this PR's** — all four flagged files are byte-identical to `feat/ui-next`, verified with `git diff --quiet origin/feat/ui-next..HEAD -- <file>`, so they pre-exist on the base and are being reported against the branch rather than introduced by it.

Judged on their merits anyway, since they are real code:

**Two are worth fixing — filed, not fixed here.**

- #355 — `logTerms.ts:283-284`, polynomial backtracking on anchored `+` quantifiers. The input is **a term the reader types into the log filter**, so it is reachable by keyboard. Consequence is a janked frame in the renderer on the user's own input, not a security boundary.
- #356 — `k8sSecret.ts:145`, the lazy `[\s\S]*?` certificate scan degrading on a bundle with many unterminated `BEGIN` markers. Narrower: needs a malformed or hostile Secret in a cluster you are already authenticated to.

Both note the same trap for whoever fixes them — a correctness-only test passes on the slow version, so the assertion has to be about work done.

**Two are false positives.** Alerts 55 and 56 flag an unescaped `.` before `kubernetes.io/last-applied-configuration` as an *"incomplete regular expression for hostnames"*. That is a Kubernetes annotation key, not a hostname, and both are in test files. Alert 57 is the same shape. Worth dismissing in the code-scanning UI as "used in tests" / "false positive" so they stop riding along on every PR — I have not dismissed them, since that is a security-alert state change and yours to make.

Not fixing any of them in this PR: they are unrelated to Helm, in files it never opens, and regex changes to `packages/core` would widen a 27-commit PR into territory a reviewer would rightly question.

---

While filing those I also made good on this PR's "follow-ups filed" line, which was a claim about work I had not actually done yet:

- #357 — dialogs are window-modal and block tab switching. **Read this one before approving:** it invalidates a wrong-cluster clearance this PR's review relied on, because that clearance rested on the modal overlay blocking the cluster rail.
- #358 — `sessions.ts` parks a terminal handle nothing closes (this PR fixed the same bug in its own copy)
- #359 — the kit's `Field` folds its hint into the control's accessible name
- #360 — the table's resize observer has never attached, so auto-sized tables never re-fit
- #361 — two suites that fail under parallelism and pass alone
- #362 — the tokens-only guard catches hex but not `rgb()` or palette classes

deveshk0:
Thanks — **both P2s are real, and I verified each in the source rather than taking them at face value.** Neither is in this PR's diff, though, so both are filed rather than fixed here.

**Why they showed up.** This PR was opened against `dev` by mistake and retargeted to `feat/ui-next` shortly after. Codex reviewed the `dev`-based diff — **420 files**, the whole accumulated new-design migration — rather than this PR's **36**. `crates/kube/src/pod_overview.rs` is not touched here; it is already on `feat/ui-next`. My error, not the reviewer's.

**Both findings confirmed:**

**#363 — name-only exclusion drops a pod in another namespace.** `fetched` is correctly keyed `(namespace, name)`, but `by_name` discards the namespace and *that* is what decides which names get queried. A `redis-0` Pending in namespace A puts `redis-0` in `by_name`, so a Running-but-crash-looping `redis-0` in namespace B is filtered out of `short`, never fetched, and absent from the Not ready list — **with `truncated` still false**, so the summary reads complete when it is not. The comment directly below the filter even says duplicate names across namespaces are expected and that the by-name fetch "brings back both of them", which is precisely what the filter defeats.

Worth flagging that this is the **same defect class this PR fixes** on the Helm screen — helm scopes a release name to a namespace, and keying by name alone let one namespace's release stand in for another's. Same mistake, different surface. Good catch.

**#364 — the cap counts names, not pods.** The budget bounds distinct names queried, but each query is cluster-wide and returns that name from every namespace holding it, with every result pushed onto `unsettled` unaccounted. One budget slot can append hundreds of bodies, overrunning `UNSETTLED_CAP` while `truncated` stays false — so the payload is both larger than intended and claims completeness. The cap bounds the wrong quantity, silently.

**Separately:** the six CodeQL alerts were also all pre-existing (files byte-identical to base). Two were worth filing — #355, #356. Three were false positives reading a Kubernetes annotation key as a hostname pattern in test fixtures; those are now dismissed with reasons.

@codex review — the diff is 36 files now, not 420.

PR #377, opened 2026-08-28
fix: 0.7.3 — download every retained pod log line, and icon-button tooltips after 200 ms

### LINKED ISSUE #353: Logs: Enable download of all Pod Logs
Currently the download of Pod logs is limited to the 5000 lines from the UI filters.

However there are use cases where you need to be able to download all stored ligs for a Pod.

### LINKED ISSUE #376: fix(ux): tooltip delay is too long
https://github.com/user-attachments/assets/7700de8b-ec04-4b4e-8a02-a342787cd0f9

I'm trying to use srelens to replace freelens as my primary k8s tool. However, since I'm not familiar with srelens yet, I rely on mouse over buttons to check the role with tooltips. The tooltip takes at least 1–2 seconds to appear, which makes exploring the UI frustrating. I'd like the tooltip delay to be shorter, especially helpful for beginners like me.

deveshk0:
Two user-reported fixes, cut from `main` so the release computes to **0.7.3** — `dev` carries `feat:` commits that would bump it to 0.8.0, which stays deferred until the new-design migration lands. Merge with a **plain merge** (not squash), then merge `main` back into `dev`.

## #353 — Logs: enable download of all pod logs

Two layers. `k8s.podLogs` forced `Some(tail_lines.unwrap_or(200))`, so no caller could ask the server for everything — omitting the field meant 200. And classic's "Download all containers" passed the view's `tailLines` and `sinceSeconds`, so "all" was capped by the filter the reporter was looking at.

- `PodLogsIn.all_lines` (default false): when set, `tail_lines: None`, which is the only thing that makes Kubernetes return every line the runtime still retains. `all_lines` wins over `tail_lines` if both are given; the field's doc says so. Every existing caller and MCP agent keeps the 200-line default. `docs/mcp-catalog.md` regenerated (asserted in sync by `srelens-registry`).
- `downloadAll()` requests `allLines: true` and drops the view's tail and since; `previous` and `timestamps` stay, since they choose *which* stream rather than how much of it.
- The button now says what it does — *"Download all container logs"*, title *"Save everything the cluster still holds for every container of these pods, not just the lines shown here"* — and does not overclaim: what log rotation has already discarded is gone.

## #376 — tooltip delay is too long

The delay was the browser's: `IconButton` rendered a native `title=`, whose ~1 s cannot be configured, and no `TooltipProvider` was mounted anywhere. `IconButton` now uses the Radix tooltip under one root provider — 200 ms, with a 300 ms skip window so sibling buttons open instantly while exploring a toolbar. **The look is unchanged by design:** the shadcn dark pill and arrow are replaced with classic's own popover surface, hairline border and small text, so the only observable difference is the delay. A disabled button wraps its trigger in a span, so a disabled reason still shows. `aria-label` is kept.

Scope held to the toolbar controls the reporter was hovering; the ~170 native `title=` attributes on other elements are a separate sweep.

## Verification

Tests first for both, then mutation: dropping the `all_lines` branch fails the Rust pin (`Some(200)` vs `None`); passing `tailLines` again fails the TS pin; restoring `title=` fails two `IconButton` pins; removing the provider fails all five; restoring the shadcn styling fails the appearance pin. Six existing assertions that read `getAttribute("title")` off icon buttons now hover and read the tooltip.

Gates: `pnpm -r test` 1254 tests, coverage 86.1% lines; `cargo test -p srelens-kube -p srelens-desktop -p srelens-registry` green; typecheck and `cargo build` clean.

Closes #353. Closes #376.

### DIFF
--- a/apps/desktop/src/components/DetailActions.test.tsx
+++ b/apps/desktop/src/components/DetailActions.test.tsx
@@ -1,5 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import { render, screen, waitFor, fireEvent } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
 import React from "react";
 
 const {
@@ -74,6 +75,14 @@ function isDisabled(el: HTMLElement): boolean {
 }
 function titleOf(el: HTMLElement): string | null {
   return el.getAttribute("title");
+}
+// IconButton explains itself through a Radix tooltip, not a native title
+// (#376): hover its trigger — the button, or the wrapper around a disabled
+// one, which is what still gets pointer events — and read the tooltip.
+async function tooltipOf(el: HTMLElement): Promise<string | null> {
+  const trigger = el.closest<HTMLElement>('[data-slot="tooltip-trigger"]') ?? el;
+  await userEvent.hover(trigger);
+  return (await screen.findByRole("tooltip")).textContent;
 }
 
 beforeEach(() => {
@@ -868,7 +877,7 @@ describe("ResourceActions", () => {
 describe("PodActions RBAC gating", () => {
   const prodPod = { ...pod, namespace: "prod" };
 
-  it("disables the Delete control and explains why when the user can't delete", () => {
+  it("disables the Delete control and explains why when the user can't delete", async () => {
     vi.mocked(useAccess).mockReturnValue({
       allowed: () => false,
       reason: () => "RBAC: no rule",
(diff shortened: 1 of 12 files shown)


deveshk0:
Superseded by https://github.com/srelens/srelens/pull/378, which targets `dev` and carries a wider #376 fix: the tooltip moved from `IconButton` to the shadcn `Button` itself, because the buttons in the reporter's screenshot (`Copy as kubectl` and the rest of that toolbar) are plain `Button`s and kept the native ~1s delay. Closing this one.

PR #378, opened 2026-08-28
fix: download every retained pod log line, and explain every button through a fast tooltip

### LINKED ISSUE #353: Logs: Enable download of all Pod Logs

### LINKED ISSUE #376: fix(ux): tooltip delay is too long

deveshk0:
Two user-reported fixes, onto `dev`. Merge `dev` → `main` afterwards to ship them; the release computes the number from the commits in range.

Supersedes #377, which carried the same two fixes based on `main` before the tooltip scope widened.

## #353 — Logs: enable download of all pod logs

Two layers, and the first meant no caller could ask for what the reporter wanted:

- `k8s.podLogs` forced `Some(tail_lines.unwrap_or(200))`, so omitting the field meant 200 rather than everything. `PodLogsIn.all_lines` (default false) now yields `tail_lines: None`, which is the only thing that makes Kubernetes return every line the runtime still retains. `all_lines` wins over `tail_lines` when both are given, and the field's doc says so. Every existing caller and MCP agent keeps the 200-line default. `docs/mcp-catalog.md` regenerated — `srelens-registry` asserts it in sync.
- Classic's "Download all containers" passed the view's `tailLines` and `sinceSeconds`, so "all" was bounded by the filter on screen. It now requests `allLines: true` and drops both; `previous` and `timestamps` stay, since they choose *which* stream rather than how much of it.

The button says what it does — *"Download all container logs"* / *"Save everything the cluster still holds for every container of these pods, not just the lines shown here"* — without overclaiming: what log rotation has discarded is gone.

## #376 — tooltip delay is too long

The delay was the browser's, and the first pass fixed too little. `IconButton` was converted, but the buttons in the screenshot are not all `IconButton`s — `Copy as kubectl` is a plain `<Button … title=…>`, and the detail toolbar has ~20 more. Those kept the native tooltip and its fixed ~1 s.

So the fix moved to the seam they already share: **`Button` renders its `title` as a Radix tooltip** instead of forwarding a native attribute, under one root provider at **100 ms** with a 300 ms skip window so sibling controls open instantly while exploring a toolbar. `IconButton` passes `title` down rather than keeping a second mechanism. Raw `<button title=…>` controls were converted; a `title` on a non-interactive element — a truncated span, a table cell — keeps the native one, which is right there.

**The appearance is deliberately unchanged.** The shadcn default (dark pill, arrow) is replaced with classic's own popover surface, hairline border and small dark text, no arrow, so the only observable difference is the delay. A pin fails if the shadcn styling comes back.

A disabled button gets no pointer events from Radix, so its trigger is wrapped in a span — the disabled reason is the only thing several of these titles carry.

## Verification

Tests first for both, then mutation: dropping the `all_lines` branch fails the Rust pin (`Some(200)` vs `None`); passing `tailLines` again fails the TS pin; forwarding `title` natively again fails the button pins; removing the provider fails all of them; restoring the shadcn styling fails the appearance pin.

Six assertions that read `getAttribute("title")` off a button now hover and read the tooltip. `tooltipOf` resolves the trigger's own `aria-describedby` rather than the first node with the tooltip role — opening a menu focuses its first item and Radix opens a tooltip on focus as well as hover, so a bare `findByRole("tooltip")` returned the first item's text however far down the menu the hover went. That one bit during this work and is worth knowing for the next such test.

Gates: 1454 frontend tests across 174 files, `cargo test -p srelens-kube -p srelens-registry` green (326 + 45), typecheck clean across all four packages.

Closes #353. Closes #376.

### DIFF
--- a/apps/desktop/src/components/AppLogView.tsx
+++ b/apps/desktop/src/components/AppLogView.tsx
@@ -3,6 +3,7 @@ import { Check, Copy, FolderOpen, RefreshCw } from "lucide-react";
 import { appLogPath, readAppLog, revealAppLog } from "@srelens/core";
 import { Button, IconButton, Select, Spinner, TextInput } from "../ui";
 import { isTauri } from "@srelens/core/platform";
+import { TitleTooltip } from "@/components/ui/tooltip";
 
 /** Log levels emitted by tauri-plugin-log, most→least severe. */
 const LEVELS = ["ERROR", "WARN", "INFO", "DEBUG", "TRACE"] as const;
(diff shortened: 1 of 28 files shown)


navneet066 (approved):
LGTM

PR #401, opened 2026-09-01
build: move the CVE pins to where pnpm 11 reads them, and pin pnpm itself

deveshk0:
pnpm 11 stopped reading the `pnpm` field in `package.json`. It warns and carries on — so a local install re-resolved without the seven advisory overrides and rewrote `pnpm-lock.yaml` with its `overrides:` block gone. CI pinned pnpm 9, which still read the field, so nothing failed and the drift was invisible.

It happened **three times** on one machine before it was understood. Each time the versions it landed on were still above the advisory floors, so nothing was exposed — but the enforcement was gone and the next resolution had nothing holding it.

## What moved

| | before | after |
|---|---|---|
| CVE overrides | `pnpm.overrides` in `package.json` — ignored by pnpm 11 | `overrides:` in `pnpm-workspace.yaml` |
| pnpm version | `version: 9` in five workflow steps; whatever is installed locally | `packageManager` in `package.json`; the five steps read it |
| esbuild's build script | `onlyBuiltDependencies` — still `ERR_PNPM_IGNORED_BUILDS` | `allowBuilds: esbuild: true` |

## Verified, not assumed

With the overrides in their new home, pnpm 11 compares them against the lockfile's own `overrides:` block, finds them identical, and **skips resolution entirely** — the lockfile is byte-for-byte unchanged and all seven pins are intact. Before the move, the same command re-resolved and dropped them. Two consecutive installs are clean and idempotent, and `pnpm test` passes with its coverage floors.

The lockfile stays at `lockfileVersion: 9.0`, so no regeneration churn lands in this diff.

## pnpm was damaging the file itself

Worth recording, because it is not obvious. When pnpm 11 has no answer about a build script it **writes its own prompt into `pnpm-workspace.yaml`** — the literal line `esbuild: set this to true or false`, a string where a boolean belongs — which then fails every later install with a YAML or approval error. That is what made `pnpm dev` fail, and it re-appeared during this work minutes after the file was cleaned. An explicit `true` is what stops it asking.

Both files now carry a comment naming the silent failure mode, because losing the `overrides:` block does not announce itself.

## One source of truth for the version

The five `pnpm/action-setup` steps lose their `version:` so the action reads `packageManager`. A second place to state the version is exactly how local and CI came to disagree.

**I could not verify locally that `pnpm/action-setup` v6 reads `packageManager` when `version:` is absent.** It is the documented behaviour and the recommended setup, but the proof is this PR's own CI run. The failure mode is loud — all five jobs fail immediately rather than shipping something wrong.

## Hard cutover

Once this merges, a contributor on pnpm 9 or 10 needs to let pnpm self-switch to the pinned version. `docs/DEVELOPMENT.md` is updated to say so.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,9 +35,11 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      # No `version:` — `pnpm/action-setup` reads `packageManager` from
+      # package.json. One source of truth: a second one here is how local
+      # ended up on pnpm 11 while CI ran 9, which silently dropped the CVE
+      # overrides from the lockfile on every local install.
       - uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
-        with:
-          version: 9
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: 22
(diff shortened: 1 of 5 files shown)


navneet066 (commented):
LGTM

navneet066 (approved):

Issue #374, opened 2026-08-27
fix(ui-next): every capability whose only consumer is classic's App.tsx


deveshk0:
`apps/desktop/src/main.tsx` mounts `App` (classic) or `NextApp` and never both. Every capability whose only consumer is an effect or a component in `apps/desktop/src/App.tsx` is therefore silently absent while the new design is running — the backend still does its half, the frontend half is just not there. Two instances of this were found independently: deep links (#370) and the MCP HTTP server (PR #373, finding 1).

This is the swept list, so the rest are known rather than found one at a time. Each row names the `App.tsx` site and what `packages/ui-next` has instead.

## Absent — no equivalent in `packages/ui-next`

1. **The MCP tool consent prompt.** `<McpConfirmDialog />` (`App.tsx:1219`) listens app-wide for `mcp://confirm-request` and answers with `respondToConfirm`. Nothing in `packages/ui-next` listens for that event. So under the new design a destructive capability call from an agent blocks in Rust, raises no prompt, and is denied when it times out — and `settings/AgentAccess.tsx` says "every change it proposes stops at a confirmation prompt", which is true about the block and false about the prompt. The assistant's inline card (`AssistantConversation.tsx:837`) is classic's too.

2. **The toast host.** `installToastNotifier()` runs unconditionally in `main.tsx`, but `<Toaster />` is mounted only at `App.tsx:1217`. `packages/ui-next` has 12 live `notify.*` call sites (`Overview.tsx`, `Forwards.tsx`, `ResourceMenu.tsx`, `settings/McpServer.tsx`, `settings/SecurityPane.tsx`, `forwards/NewForwardDialog.tsx`, `lib/sessions.ts`) — every one of them creates a toast with nothing to render it. Two files already work around this (`index.tsx:93`, `settings/AppearancePane.tsx:175`) by rendering their own error surface, which is the right shape locally and does not help the other twelve.

3. **`kubeconfig-changed`.** `App.tsx:393` re-lists contexts when the backend reports the kubeconfig changed. `lib/clusters.ts` is written only at boot by `shell/Window.tsx:217` and by the explicit `Refresh` on `/connections` and `/connect`, so an external edit — `kubectl config use-context`, a file added outside srelens — leaves the new design's cluster list stale until a reader presses Refresh.

[...]

PR #413, opened 2026-09-01
fix(ui): confirm a copy on the button that did it

### LINKED ISSUE #410: "Copy as kubectl" gives no confirmation — its success toast reaches nobody in the new design
### What happens

Clicking **Copy as kubectl** in the detail panel's footer bar copies the command
and then looks exactly as it did before. Nothing confirms it worked, so the
natural response is to click it again, and again — with no way to tell whether
any of them landed.

### Why — it is not that nobody wrote the feedback

The handler already reports success. `copyKubectlCommand`
(`packages/core/src/lib/copyKubectl.ts:3-11`) writes to the clipboard and calls
`notify.success("Copied kubectl command")`.

**That toast is never shown in the new design.** The only toast host in the app
is sonner's `<Toaster>`, mounted inside the classic tree at
`apps/desktop/src/App.tsx:1217`; `NextApp` renders no equivalent. The new design
already knows this and says so in three separate places:

- `packages/ui-next/src/shell/Window.tsx:340-342` — "The toast host lives in the
  classic tree … so `notify` from here reaches nobody."
- `packages/ui-next/src/shell/Window.tsx:409-411` — "`notify` reaches nobody in
  this tree, which is the same gap `lockNow` names."
- `packages/ui-next/src/shell/AgentConsent.tsx:49-58` — "`notify`'s sink is
  installed at `main.tsx` for both designs, but it calls sonner, and sonner's
  `<Toaster>` is mounted in classic's `App` — which the new design does not
  render."

(The kit's own `Toast` is presentational only — it renders a toast it is handed,
with no host, queue or portal, `packages/ui-kit/src/Toast.tsx:55-59` — and
nothing in `ui-next` mounts it.)

So the button believes it is confirming and the reader sees silence.

> **The wider gap is already tracked.** #374 item 2, "The toast host", records
> the same thing and its full extent: `installToastNotifier()` runs
> unconditionally in `main.tsx` while `<Toaster />` is mounted only at
> `App.tsx:1217`, leaving twelve live `notify.*` call sites in `packages/ui-next`
> with nothing to render them. This request neither duplicates that nor waits for
> it — the on-button confirmation below needs no toast host at all, and is the
> better affordance for a copy button regardless of how #374 is resolved.

On top of that, a clipboard *failure* is swallowed without a word
(`copyKubectl.ts:8-10`), so an unsuccessful copy is invisible twice over.

### What to do instead

Put the confirmation **on the button**. It is what the reader is looking at, it
needs no host, and it survives the toast gap entirely.

```
   default            just clicked           after ~1.4s
┌────────────────┐  ┌────────────────┐    ┌────────────────┐
│ ⧉ Copy as      │  │ ✓ Copied       │ →  │ ⧉ Copy as      │
│   kubectl      │  │                │    │   kubectl      │
└────────────────┘  └────────────────┘    └────────────────┘
                      check + word,
                      then reverts
```

### The pattern already exists — reuse it, do not invent one

`CopyCommand` (`packages/ui-kit/src/CopyCommand.tsx:81-110`) already does
exactly this, and its rationale is written down at `:75-79`:

> The confirmation is a word AND an ok-toned check, and it reverts after 1.4s. A
> label that never comes back reads as a control already spent; the check is what
> makes "Copied" land as confirmation rather than as the button's new name.

Its shape, which this should match:

- `const [copied, setCopied] = useState(false)` (`:82`).
- A `useEffect` timer that reverts after `COPIED_MS = 1400` (`:6`, `:84-88`) and
  **clears on unmount** via the effect's cleanup. Re-clicking re-runs the effect,
  so a rapid second click restarts the window rather than stacking timers or
  sticking.
- A `try/catch` where the catch deliberately does **not** set `copied`
  (`:94-98`): *"Saying 'Copied' when nothing was copied would be the only real
  harm."* That rule is already house style and this button must follow it.

**But the component cannot be dropped in as-is.** `CopyCommand` renders the
command text in a `<code>` beside its button (`:102-107`); the footer entry is a
bare action with no room for the command. So the *pattern* is reusable and the
*component* is not — which is the actual work in this issue:

- Lift the copied-state logic out of `CopyCommand` into something shared (a
  `useCopied()` hook, or a `CopyButton` that `CopyCommand` then uses), so there
  is one timer, one failure rule and one revert delay in the codebase.
- The bar renders generic entries: `DetailActions` builds `ActionBarAction`s
  from the row menu's items (`packages/ui-next/src/screens/detail/DetailActions.tsx:65-76`),
  and `ActionBarAction` (`packages/ui-kit/src/ActionBar.tsx:9-23`) has `id`,
  `label`, `icon`, `danger`, `disabledReason`, `onSelect` — **no transient
  confirmed state**. Either that type gains a way to show one, or the copy entry
  supplies its own control. `Button` itself is not the place: `ButtonProps`
  (`packages/ui-kit/src/Button.tsx:16-19`) is `variant` + `size` over the native
  props, and a transient copied state is not a button variant.

### Fix it centrally — there are four silent copy affordances

The detail footer and the row menu are the same call site
(`packages/ui-next/src/screens/ResourceMenu.tsx:386`, reached from both hosts),
but they are not the only ones:

| Where | Line | Feedback today |
| --- | --- | --- |
| Row menu / detail footer "Copy as kubectl" | `ui-next/src/screens/ResourceMenu.tsx:386` | toast that never renders |
| Overview's copy action | `ui-next/src/screens/Overview.tsx:1056` | toast that never renders |
| `KubectlPreview`'s copy icon button | `ui-kit/src/KubectlPreview.tsx:56` | none at all |
| Forwards' address copy | `ui-next/src/screens/Forwards.tsx:352-356` | toast that never renders; failure swallowed |

A shared hook or `CopyButton` fixes all four; a local `useState` in
`DetailActions` fixes one and leaves three.

### Accessibility

A visual glyph swap says nothing to a screen reader. Two things are needed:

- **Announce it.** "Copied to clipboard" through a polite live region. The kit
  has no announcer helper today — `LiveSignal`, `StatusPill` and `StatusBar` are
  status *displays*, and the only `aria-live` in the kit is `ConsoleDock.tsx:192`
  — so this is a small new piece. Note `Toast` chooses its live region by tone
  (`Toast.tsx:61-65`); the same reasoning applies: a copy confirmation is
  `status`, not `alert`.
- **Keep the accessible name honest.** `CopyCommand` deliberately has no
  `aria-label` over its visible text (`:78-79`). If the visible label changes to
  "Copied", the accessible name changes with it, which is correct — but the
  *announcement* should not rely on that alone, since a name change on a button
  the user just activated is not reliably spoken.

Reverting must not animate for a reader who has asked it not to
(`prefers-reduced-motion`).

### Acceptance criteria

- [ ] Clicking **Copy as kubectl** swaps the icon to a check and the label to
      "Copied", reverting after ~1.4 s — matching `CopyCommand`'s existing timing
      and shape.
- [ ] The confirmation appears with no toast host present, i.e. it works in the
      new design as it stands today.
- [ ] A clipboard write that rejects shows **no** "Copied" — and is not silent
      either; the reader is told it failed.
- [ ] Clicking again during the confirmation restarts the window; no flicker, no
      stuck "Copied", no stacked timers.
- [ ] The timer is cleared on unmount (closing the peek mid-confirmation must not
      warn or leak).
- [ ] "Copied to clipboard" is announced politely to assistive technology.
- [ ] The revert respects `prefers-reduced-motion`.
- [ ] The confirmation lives in shared code, so the detail footer, the row menu,
      Overview and `KubectlPreview` all get it — with one copy of the timer and
      the failure rule.
- [ ] `CopyCommand` still behaves exactly as it does now, on the same timing.

### Environment

- srelens **0.8.0** (`apps/desktop/src-tauri/tauri.conf.json:5`). Repo at `dev`
  / `9abb2f9`.
- New design (`packages/ui-next` on `packages/ui-kit`).

navneet066:
Closes #410.

"Copy as kubectl" reported success through `notify.success`, and the new design mounts no toast host (#374 item 2) — so every copy was silent. The reader clicked, nothing happened, and they clicked again. Four copy affordances were in that state, and a fifth, `CopyCommand`, already had the confirmation the others needed, written into itself.

## What this does

Lifts that logic into a shared `useCopied` hook: one delay, one timer cleared on unmount, one rule about failure. On top of it, two small pieces — `CopyIconButton` for the icon-only places, and `CopyAnnounce` for the spoken half.

- **On-button confirmation.** The glyph becomes a check and the word becomes "Copied", reverting after 1.4s — the timing and shape `CopyCommand` already used.
- **A failed copy never says "Copied".** It reports failure both when the copy throws and when it resolves `false`, so a helper that reports by return value is not read as a success.
- **Rapid re-click restarts the window.** The state is held in an object rather than as a bare `CopyState`: clicking again while "Copied" is up records an identical value, React bails out of the update, the effect never re-runs, and the second confirmation would expire on the first click's schedule. A fresh object is never equal to the last, so the effect re-runs and the window restarts.
- **One announcement per control.** A control with a visible word confirms in that word, and its accessible name follows the word. `CopyAnnounce` — a `role="status"` region, a sibling and never a child — carries the outcome for the icon-only controls, which have no word to change and whose name therefore stays the action.
- **Reduced motion respected.** The confirmation tint's transition is dropped under `prefers-reduced-motion`.

`copyKubectlCommand` now returns whether it worked. It used to swallow a denied clipboard, which left every caller unable to tell a copy that happened from one that did not. Classic's toast is untouched.

## Where it is wired

| Site | |
| --- | --- |
| "Copy as kubectl" — detail footer, on the bar **and in the overflow** | via a new optional `ActionBarAction.confirmLabel` |
| Overview's node copy action | same |
| `KubectlPreview` | also gives the classic design the confirmation, free |
| Forwards — both row controls | kubectl command and address |
| `CopyCommand` | unchanged behaviour, now from the shared hook |
| The list's own right-click `ContextMenu` | **not** wired — it carries `confirmLabel` to `DetailActions` but draws no confirmation of its own |

## One behaviour change to shipped code

**A failed copy now says so.** It used to stay silent, which is indistinguishable from a copy that worked — the whole complaint in #410. A bar action reads "<action> failed" — "Copy as kubectl failed" — and reverts; `CopyCommand` reads "Copy failed".

`CopyCommand`'s test had pinned the silent behaviour ("still reads Copy" after a rejected write), so it is updated: it keeps its original assertions — no false claim, command still readable — and adds the new one.

## Review round (@deveshk0)

Both blockers and all three smaller points are addressed in `f573135`.

- **`copyKubectl.ts` returned `true` with no clipboard.** `await navigator.clipboard?.writeText(...)` resolves `undefined`, so the happy path ran on to `return true` and raised the toast having copied nothing — every non-secure origin. Now checked rather than optional-chained, matching `CopyCommand` and `CopyIconButton`. The "denied/unavailable" test was only mocking a rejecting `writeText`; the clipboard-absent half is now its own case.
- **`confirmLabel` reached the bar only.** Whether an action is on the bar depends on how many come before it, so Copy as kubectl confirmed on a ConfigMap and not on a Pod, in the same footer. Fixed by the reviewer's first route: the overflow holds still for a confirming pick and closes when the confirmation is spent, so the answer arrives wherever the action landed. Pinning it to the bar was the alternative and was rejected — it would put a third button in a 260px footer the design draws with two, and it would fix one action rather than the rule. Verified against the real Pod peek path in `ResourceDetailView.test.tsx`, not only against `ActionBar` in isolation.
- **The `seq` counter was not doing what its comment said.** Correct, and reproduced: freezing the counter leaves the suite green, so it was decorative. It is gone, and the comment now names the object identity that actually does the work. Verified by mutation — collapsing the state to a bare `useState<CopyState>` fails "restarts the window on a rapid second copy rather than expiring early", which is the guard the comment points at.
- **The outcome was announced twice.** Fixed toward `CopyButton`'s split, and the deciding reason is WCAG 2.5.3: a button reading "Copied" under a pinned name of "Copy as kubectl" has a visible word missing from its own accessible name, which strands anyone driving the app by voice. So the name is pinned only where there is no visible word — the icon-only controls, which is exactly `CopyButton`'s `iconOnly` branch — and the controls with a word carry the outcome in it, once. `CopyCommand` lost the live region it had been given.
- **The description said "Copy failed".** Corrected above; the code's wording stands.

On the #392 collision: agreed, and `CopyButton` is deliberately untouched here so it can be rebased onto `useCopied` after this lands.

## Verification

- `pnpm -r typecheck` — clean across all four packages.
- `pnpm test` (root, with coverage — the real gate) — **5483 passing, 0 failing**; coverage 92.91% lines against the 85% gate. `useCopied.ts`, `copyKubectl.ts`, `CopyCommand.tsx`, `CopyAnnounce.tsx` and `ActionBar.tsx` are each at 100%.
- `cargo test --workspace` — 1233 passing, 0 failing at the previous commit (no Rust touched; run because CONTRIBUTING gates on it).

New tests since the review: the clipboard-absent case; confirming from the overflow, the menu staying open for it and closing when the confirmation is spent, and refusing to say "Copied" there on a failure; the Pod peek footer end to end; and `CopyIconButton`'s own file — the name that does not change, the announcement, the failure in both channels, and nothing left announcing afterwards.

### DIFF
--- a/packages/core/src/lib/copyKubectl.ts
+++ b/packages/core/src/lib/copyKubectl.ts
@@ -1,11 +1,24 @@
 import { notify } from "./notify";
 
-/** Copy a kubectl command to the clipboard, silently ignoring a denied/unavailable clipboard (matching ResourceOverview's copy affordance). */
-export async function copyKubectlCommand(command: string): Promise<void> {
+/**
+ * Copy a kubectl command to the clipboard.
+ *
+ * Returns whether it worked, and that return value is the point: a denied or
+ * absent clipboard used to be swallowed here without a word, so a copy that
+ * never happened was indistinguishable from one that did. The callers that
+ * draw a confirmation need to know which they got — saying "Copied" when
+ * nothing was copied is the one outcome worse than saying nothing. (#410)
+ *
+ * The toast stays for the classic design, whose `<Toaster>` renders it. Under
+ * the new design it reaches nobody (#374 item 2), which is why the copy
+ * controls there confirm on the button instead and do not rely on this.
+ */
+export async function copyKubectlCommand(command: string): Promise<boolean> {
   try {
     await navigator.clipboard?.writeText(command);
     notify.success("Copied kubectl command");
+    return true;
   } catch {
-    /* clipboard unavailable — ignore, matching ResourceOverview's copy affordance */
+    return false;
   }
 }
--- a/packages/core/src/lib/copyKubectl.ts
+++ b/packages/core/src/lib/copyKubectl.ts
@@ -14,8 +14,17 @@ import { notify } from "./notify";
  * controls there confirm on the button instead and do not rely on this.
  */
 export async function copyKubectlCommand(command: string): Promise<boolean> {
+  // Checked, not optional-chained. `await navigator.clipboard?.writeText(...)`
+  // resolves `undefined` when there is no clipboard at all, so the happy path
+  // ran on to `return true` and raised the success toast having copied nothing
+  // — the exact "Copied" over an empty clipboard this function exists to stop.
+  // `navigator.clipboard` is absent on any non-secure origin, so web mode
+  // served over plain http to anything but localhost hit it every time.
+  // `CopyCommand` and `CopyIconButton` call `writeText` unguarded and so report
+  // the failure correctly; this now agrees with them.
+  if (!navigator.clipboard) return false;
   try {
-    await navigator.clipboard?.writeText(command);
+    await navigator.clipboard.writeText(command);
     notify.success("Copied kubectl command");
     return true;
   } catch {
--- a/packages/ui-kit/src/ActionBar.tsx
+++ b/packages/ui-kit/src/ActionBar.tsx
@@ -1,10 +1,12 @@
-import type { ReactNode } from "react";
+import { useState, type ReactNode } from "react";
 import { Button } from "./Button";
+import { CopyAnnounce } from "./CopyAnnounce";
 import { cx } from "./cx";
 import type { IconComponent } from "./IconButton";
 import { Popover } from "./Popover";
 import { filled } from "./slot";
 import { toneColor } from "./tone";
+import { useCopied } from "./useCopied";
 
 export interface ActionBarAction {
   /** Unique within the bar. Identifies the action, never shown. */
@@ -19,7 +21,17 @@ export interface ActionBarAction {
    * a blocked action with no explanation is worse than no action at all.
    */
   disabledReason?: string;
-  onSelect: () => void;
+  /**
+   * Turns this into an action that answers. After `onSelect` resolves — and
+   * does not resolve `false` — the button shows this word with a check for a
+   * moment, then goes back to `label`.
+   *
+   * Only the bar draws it. A menu row closes on the pick, so a confirmation
+   * there would be drawn on something already gone; the point of this is the
+   * button the reader is still looking at. (#410)
+   */
+  confirmLabel?: string;
+  onSelect: () => void | boolean | Promise<void | boolean>;
 }
 
 export interface ActionBarProps {
@@ -108,29 +120,7 @@ export function ActionBar({
       aria-label={label}
       className={cx("flex flex-wrap items-center gap-1.5", className)}
     >
-      {head.map((a) => {
-        const blocked = filled(a.disabledReason);
-        const Icon = a.icon;
-        return (
-          <Button
-            key={a.id}
-            type="button"
-            variant={a.danger ? "danger" : "secondary"}
-            // Not `disabled`: see the note above — the reason has to stay
-            // reachable, and a disabled button cannot be focused to read it.
-            aria-disabled={blocked || undefined}
-            title={a.disabledReason}
-            className="aria-disabled:cursor-not-allowed aria-disabled:opacity-50"
-            onClick={() => {
-              if (blocked) return;
-              a.onSelect();
-            }}
-          >
-            {Icon && <Icon size={12} aria-hidden="true" />}
-            {a.label}
-          </Button>
-        );
-      })}
+      {head.map((a) => <BarButton key={a.id} action={a} />)}
 
       {rest.length > 0 && (
         <Popover
@@ -196,5 +186,68 @@ export function ActionBar({
         </Popover>
       )}
     </div>
+  );
+}
+
+/* Inline rather than an icon-set import: the kit takes no dependency on lucide,
+   and this is the only glyph it needs. */
+function CheckGlyph() {
+  return (
+    <svg width="12" height="12" viewBox="0 0 24 24" fill="none" aria-hidden="true">
+      <path
+        d="m5 13 4 4 10-10"
+        stroke="currentColor"
+        strokeWidth="3"
+        strokeLinecap="round"
+        strokeLinejoin="round"
+      />
+    </svg>
+  );
+}
+
+/**
+ * One action on the bar.
+ *
+ * A component of its own rather than JSX inside the map, because an action that
+ * confirms needs state and a hook cannot live in a loop — the number of actions
+ * changes between renders, which is the same reason `Rail` reads its probes once
+ * for the whole list. Each button owning its own confirmation is also what keeps
+ * two of them from sharing one: copying and then restarting must not flash
+ * "Copied" on Restart.
+ */
+function BarButton({ action: a }: { action: ActionBarAction }) {
+  const { state, run } = useCopied();
+  const blocked = filled(a.disabledReason);
+  const confirming = filled(a.confirmLabel) && state === "copied";
+  const failed = filled(a.confirmLabel) && state === "failed";
+  const Icon = a.icon;
+
+  return (
+    <>
+      <Button
+        type="button"
+        variant={a.danger ? "danger" : "secondary"}
+        // Not `disabled`: see the note above — the reason has to stay
+        // reachable, and a disabled button cannot be focused to read it.
+        aria-disabled={blocked || undefined}
+        title={a.disabledReason}
+        className="aria-disabled:cursor-not-allowed aria-disabled:opacity-50"
+        onClick={() => {
+          if (blocked) return;
+          // Only an action that asked to confirm is awaited for its answer;
+          // every other one keeps the fire-and-forget call it had.
+          if (filled(a.confirmLabel)) void run(a.onSelect);
+          else void a.onSelect();
+        }}
+      >
+        {confirming ? <CheckGlyph /> : Icon && <Icon size={12} aria-hidden="true" />}
+        {confirming ? a.confirmLabel : failed ? `${a.label} failed` : a.label}
+      </Button>
+      {/* Outside the button, not inside it. A live region nested in the control
+          becomes part of the control's accessible name, so the button would be
+          called "Copied Copied to clipboard" — the announcement read back as
+          the name of the thing that made it. */}
+      {filled(a.confirmLabel) && <CopyAnnounce state={state} />}
+    </>
   );
 }
--- a/packages/ui-kit/src/ActionBar.tsx
+++ b/packages/ui-kit/src/ActionBar.tsx
@@ -1,12 +1,11 @@
-import { useState, type ReactNode } from "react";
+import { useEffect, useRef, useState, type ReactNode } from "react";
 import { Button } from "./Button";
-import { CopyAnnounce } from "./CopyAnnounce";
 import { cx } from "./cx";
 import type { IconComponent } from "./IconButton";
 import { Popover } from "./Popover";
 import { filled } from "./slot";
 import { toneColor } from "./tone";
-import { useCopied } from "./useCopied";
+import { useCopied, type CopyState } from "./useCopied";
 
 export interface ActionBarAction {
   /** Unique within the bar. Identifies the action, never shown. */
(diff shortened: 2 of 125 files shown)


packages/core/src/lib/copyKubectl.ts
```diff
@@ -1,11 +1,24 @@
...
+ * draw a confirmation need to know which they got — saying "Copied" when
+ * nothing was copied is the one outcome worse than saying nothing. (#410)
+ *
+ * The toast stays for the classic design, whose `<Toaster>` renders it. Under
+ * the new design it reaches nobody (#374 item 2), which is why the copy
+ * controls there confirm on the button instead and do not rely on this.
+ */
+export async function copyKubectlCommand(command: string): Promise<boolean> {
   try {
     await navigator.clipboard?.writeText(command);
```
deveshk0: `navigator.clipboard?.writeText(...)` is `undefined` when there's no clipboard, so the `await` resolves and this returns `true`. It fires `notify.success` too. I ran it with `navigator.clipboard` set to undefined and got `true`.

That's the "Copied" with nothing copied case. `navigator.clipboard` is undefined on non-secure origins, so web mode served over plain http to anything but localhost hits it.

`CopyCommand` and `CopyIconButton` call `writeText` without the `?.`, so they throw and report failure correctly. Suggest matching them:

```ts
if (!navigator.clipboard) return false;
```

The test at `copyKubectl.test.ts:28` is titled "denied/unavailable" but only mocks a rejecting `writeText`, so the unavailable half isn't covered.

packages/ui-kit/src/ActionBar.tsx
```diff
@@ -19,7 +21,17 @@ export interface ActionBarAction {
...
+  /**
+   * Turns this into an action that answers. After `onSelect` resolves — and
+   * does not resolve `false` — the button shows this word with a check for a
+   * moment, then goes back to `label`.
+   *
+   * Only the bar draws it. A menu row closes on the pick, so a confirmation
+   * there would be drawn on something already gone; the point of this is the
+   * button the reader is still looking at. (#410)
+   */
+  confirmLabel?: string;
```
deveshk0, line 37: `confirmLabel` is only read by `BarButton`, which renders `head`. The overflow rows still do `close(); a.onSelect();`, so nothing confirms behind the popover.

`DetailActions` passes max 2 in the peek footer and 4 in the header, and its own doc lists the order as Follow logs, Open shell, Port forward, Edit, Copy as kubectl. Copy is 5th, so it's in the overflow in both hosts. #410 is about the footer, so for a Pod the click it reports still confirms nothing.

It does work for kinds with fewer actions ahead of it (ConfigMap has no logs/shell/forward) and when RBAC filters those out, so it's inconsistent by kind rather than dead.

I haven't run the app — this is off `ON_BAR = 2` and that doc comment, so worth checking against a real Pod peek. Two ways out: confirm in the overflow, which needs the popover to stay open, or let a caller pin an action to the bar regardless of index.

Also, the description's table lists the row menu as wired through `confirmLabel`, but the doc you added to `ContextMenu` says a menu ignores it.

packages/ui-kit/src/useCopied.ts
```diff
@@ -0,0 +1,71 @@
...
+  async function run(copy: () => unknown | Promise<unknown>): Promise<void> {
+    let state: CopyState;
+    try {
+      // `false` is a refusal, not a value: a helper that reports failure by
+      // returning rather than throwing must not be read as having worked.
+      state = (await copy()) === false ? "failed" : "copied";
+    } catch {
+      state = "failed";
+    }
+    setResult((r) => ({ state, seq: r.seq + 1 }));
```
deveshk0: I changed this to `seq: r.seq` and the rapid re-click test still passed. `setResult` returns a new object literal either way, so React never bails out and the effect re-runs whether or not the counter moved.

The hazard is real though. I tried the bare-string version the comment describes (`useState<CopyState>`, `setState(next)`) and the test fails with `expected 'idle' to be 'copied'`. So it's the object identity doing the work, not the counter.

Both directions are unguarded: swap the object for a bare `CopyState` and the bug comes back with the suite green, or drop `seq` and nothing breaks. Either remove it and point the comment at the object identity, or keep it and add a test that fails without it.

deveshk0 (changes requested):
The shape is right. The hook is the correct seam, the failure rule is stated plainly, and the doc on `CopyAnnounce` explaining why it mounts on demand, and that this is a compromise, is the right thing to write down. Updating the `CopyCommand` test instead of leaving it pinning the silent behaviour is right too.

Two blockers:

- `copyKubectl.ts:18` returns `true` when there's no clipboard at all.
- `ActionBar.tsx:33` — `confirmLabel` only reaches bar actions, and Copy as kubectl isn't one in the detail footer, which is the surface #410 reports.

Smaller things:

- `useCopied.ts:67` — the sequence counter isn't doing what the comment says it does.
- `confirmLabel` renames the button and `CopyAnnounce` announces, so a screen reader can get the outcome twice, and the name of the focused control changes under the reader's own click. `CopyButton` in #392 went the other way: name stays the action, outcome announced separately. Either is defensible; they should match.
- `${a.label} failed` renders "Copy as kubectl failed", not "Copy failed" as the description says. The code reads better, so fix the description.

One collision: #392 adds `ui-kit/CopyButton.tsx`, an icon-capable copy control with its own 1.4s revert and its own `role="status"` sibling. Same job as `CopyIconButton` + `CopyAnnounce` + `useCopied`. Neither PR can see the other. Yours is the better factoring, so land this first and I'll rebase `CopyButton` onto `useCopied` and drop its own timer.

I didn't re-run the suite on this branch, so the 5474/1233 figures and the coverage numbers are as reported, not verified.

deveshk0:
Correction to the coordination note in my review: #392 merged first, so the ordering I suggested there is inverted.

`dev` now has `packages/ui-kit/src/CopyButton.tsx` — icon-capable copy control, own 1.4s revert, own `role="status"` sibling. So this branch will land beside it rather than the other way round, and ui-kit would end up with two copy controls and two confirmation mechanisms.

The factoring here is still the better one, so I'd rather this PR won: `useCopied` + `CopyAnnounce` as the shared pieces, and `CopyButton` reduced to a caller of them (or deleted in favour of `CopyIconButton`, if its `iconOnly` prop turns out to be the same thing).

That's cleanup on my side, not yours. Happy to do it as a follow-up once this lands, or to push the unification onto this branch if you'd rather it merge as one piece — your call.

One thing worth knowing while you're in this code: `CopyButton`'s live region deliberately kept the button's accessible name fixed at the action and announced the outcome separately, which is the opposite of what `confirmLabel` does here. That's the a11y point in my review, and now both behaviours are in the same package.

deveshk0 (approved):

PR #414, opened 2026-09-01
fix(kube): resolve apply scope from the kind, not the document (#404)

### LINKED ISSUE #404: fix(kube): applying a manifest without `metadata.namespace` fails with "the server could not find the requested resource"
## Summary

Any manifest for a **namespaced** kind that does not carry an explicit `metadata.namespace` fails to apply. The New Resource tab shows an inline error marker on line 1 and a toast:

> **Failed to apply Secret/test-secret**
> the server could not find the requested resource

The manifest is valid Kubernetes — `kubectl apply -f` accepts the identical file and places the object in the context's current namespace. srelens instead builds a **cluster-scoped** request path for a namespaced resource, which matches no route on the API server, so the apiserver answers `404 NotFound: the server could not find the requested resource`.

This is not specific to Secrets or to `kubernetes.io/tls`. It reproduces with any namespaced kind (ConfigMap, Deployment, Service, …) whenever `metadata.namespace` is absent.

## Environment

| | |
|---|---|
| srelens | `0.8.0` (`dev` @ `9abb2f9`) |
| Kubernetes | server v1.35.0 (local `kind` cluster) |
| Platform | macOS (Apple silicon) |

## Steps to reproduce

1. Open a cluster, then **New Resource**.
2. Pick **Template → Secret** and replace the body with a TLS Secret that has **no `namespace:`** under `metadata`:

   ```yaml
   apiVersion: v1
   kind: Secret
   metadata:
     name: test-secret
   data:
     tls.crt: <base64 cert>
     tls.key: <base64 key>
   type: kubernetes.io/tls
   ```

   To generate a throwaway pair for the repro:

   ```sh
   openssl req -x509 -newkey rsa:2048 -nodes -days 1 \
     -keyout tls.key -out tls.crt -subj "/CN=example.invalid"
   kubectl create secret tls test-secret --cert=tls.crt --key=tls.key \
     --dry-run=client -o yaml            # then delete the namespace: line
   ```

3. Click **Create**.

**Minimal repro** — no certificate needed, the same failure:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: my-config
data:
  key: value
```

### Expected

The resource is created in the namespace the tab is scoped to (or the context's default namespace), matching `kubectl apply -f` behaviour.

### Actual

Apply fails. An inline lint marker appears on line 1 with `the server could not find the requested resource`, and a toast reads `Failed to apply Secret/test-secret`. Nothing is created.

## Screenshots

New Resource tab: inline lint marker reading "the server could not find the requested resource"
*Inline lint marker on line 1 — this one comes from `k8s.validateManifest`, and appears before Create is even clicked.*

Toast reading "Failed to apply Secret/test-secret — the server could not find the requested resource"
*The apply toast after clicking Create.*

> The base64 `tls.crt` / `tls.key` values are painted out in both images; the manifest above regenerates the failure with a throwaway certificate.

## Root cause

`crates/kube/src/manifest.rs:509-515` — the apply path decides request scope from **whether the user typed a namespace**, not from whether the **kind** is namespaced:

```rust
let (group, version) = parse_api_version(&r.api_version);
let ar = ApiResource::from_gvk(&GroupVersionKind::gvk(&group, &version, &r.kind));
let api: Api<DynamicObject> = match &r.namespace {
    Some(ns) => Api::namespaced_with(client.clone(), ns, &ar),
    None => Api::all_with(client.clone(), &ar),   // <-- cluster-scoped
};
```

`r.namespace` is `metadata.namespace` read straight off the document (`resource_ref`, `crates/kube/src/manifest.rs:26-39`) and is `None` when the field is absent. `Api::all_with` builds the collection path `/api/v1/secrets`, so the subsequent `api.patch(&r.name, …)` (`crates/kube/src/manifest.rs:522`) targets:

```
PATCH /api/v1/secrets/test-secret          # what srelens sends
PATCH /api/v1/namespaces/<ns>/secrets/test-secret   # what it should send
```

There is no `…/secrets/{name}` route at cluster scope, so the apiserver 404s with exactly the reported string. Confirmed against a live cluster:

```console
$ kubectl get --raw '/api/v1/secrets/test-secret'
Error from server (NotFound): the server could not find the requested resource   # <-- the reported error

$ kubectl get --raw '/api/v1/namespaces/default/secrets/test-secret'
Error from server (NotFound): secrets "test-secret" not found                    # correct path, normal 404
```

The same holds for other namespaced kinds — `/api/v1/configmaps/my-config` and `/apis/apps/v1/deployments/my-app` both return the routing 404.

### The read paths already get this right

`k8s.getManifest` and `k8s.getObject` resolve scope from the **kind**, via the `namespaced` flag in `gvk_for` (`crates/kube/src/manifest.rs:237-285`), and fall back to `default` when no namespace is supplied:

```rust
// crates/kube/src/manifest.rs:185-190  (and :301-306)
let api: Api<DynamicObject> = if namespaced {
    let ns = input.namespace.as_deref().filter(|s| !s.is_empty()).unwrap_or("default");
    Api::namespaced_with(client, ns, &ar)
} else {
    Api::all_with(client, &ar)
};
```

The three manifest **write/dry-run** paths do not do this.

### Why editing an existing resource still works

The Edit tab seeds its YAML from `k8s.getManifest`, which returns the live object — and a live namespaced object always carries `metadata.namespace`. So `r.namespace` is `Some(...)` and the correct path is built. The bug only surfaces for hand-written or pasted manifests, which is exactly the New Resource flow.

### Why the built-in templates hide it

Every namespaced starter template in `apps/desktop/src/components/NewResourceEditor.tsx:7-84` interpolates `namespace: ${ns}`, so an unedited template applies fine. The failure appears as soon as a user pastes their own manifest or removes that line — and manifests copied from upstream docs, Helm output or `kubectl create --dry-run` frequently have no namespace.

## Blast radius

The same `match &r.namespace` shape appears in three capabilities, all in `crates/kube/src/manifest.rs`:

| Capability | Line | User-visible effect |
|---|---|---|
| `k8s.applyManifest` | 513-514 | Create/apply fails with the 404 toast |
| `k8s.validateManifest` (`validate_document`) | 596-597 | Spurious inline lint error on a valid manifest |
| `k8s.diffManifest` | 827-828 | Changes panel diffs against the wrong path — the object reads as non-existent, so the diff shows a full create |

`k8s.applyManifest` is also exposed through the built-in MCP server (`crates/mcp/src/lib.rs:285`), so agent-authored manifests hit the same wall.

(`crates/kube/src/watch.rs:835` uses the same shape but is a list/watch, where all-namespaces is the correct meaning.)

## Suggested direction

Resolve scope from the kind rather than from the presence of the field — reuse the `gvk_for(...).1` / `resolve_api_resource` logic the read paths already use, and fall back to the namespace the request is scoped to when the document omits one. Note `ApplyIn` (`crates/kube/src/manifest.rs:385-392`) currently carries no `namespace`, so a fallback namespace has to be threaded in from the caller (`NewResourceEditor` already receives one as a prop) for the fallback to honour the tab's namespace rather than hardcoding `default`.

A regression test at the `Api` path level for a namespaced kind with no `metadata.namespace` would lock this down for all three capabilities.

navneet066:
Closes #404.

## The bug

Applying a manifest for a **namespaced** kind that omits `metadata.namespace` failed with `the server could not find the requested resource`. The apply path chose the request scope from *whether the document named a namespace*, not from *whether the kind is namespaced* — so a namespaceless Secret was sent to the cluster-scoped path `PATCH /api/v1/secrets/test-secret`, which matches no route, and the API server answered 404.

## The fix

A single shared helper `apply_namespace(kind, group, version, doc_ns, fallback)` resolves the namespace to use, and all three write/dry-run paths now go through it:

- `applyManifest`
- `validateManifest` (`validate_document`) — the source of the spurious inline lint marker
- `diffManifest` — which now also reports the namespace it *actually* diffed against, rather than the one the document happened to name (they differ exactly when the document named none)

`ApplyIn` / `ValidateIn` / `DiffIn` each gain an optional `namespace`, threaded from the caller (`ManifestEditor` → `NewResourceEditor` / `YamlView` / `EditResourceTab`), so a New-resource tab scoped to `staging` creates in `staging`, not silently in `default`.

### Scope rules

- **Namespaced kind** → the document's namespace, else the caller's, else `default` (matching the read paths' existing last resort).
- **Cluster-scoped kind** → no namespace, even if the document names one (the API server ignores it there).
- **Kind unknown to `gvk_for` (a CRD)** → **behaviour unchanged, deliberately.** Nothing in-process knows a CRD's scope without asking the cluster, and defaulting an unknown kind to namespaced would break applying a cluster-scoped CRD (e.g. a `ClusterIssuer`) that works today. This trades no known bug for an unknown regression.
- `gvk_for` keys on the kind alone, so its answer is trusted only when the document's **group and version** also match the entry — otherwise a custom `Service` in another group could be forced into a namespace it may not have. Covered by a test.

`watch.rs` (the list/watch path, where all-namespaces is correct) is untouched.

## Verification

- **The regression test is proven to fail pre-fix:** reverting `validate_document` to the old behaviour makes it fail with the exact URL from the issue (`PATCH /api/v1/secrets/test-secret`, no namespace); restoring the fix makes it pass.
- **Root `pnpm test` (with coverage): 5464 tests pass, 92.89% lines** (over the 85% gate). `cargo test --workspace`: all pass.
- **7 new tests** — 4 unit tests exhaustively covering `apply_namespace`'s branches (namespaceless→default, caller-wins-over-default, explicit-wins, empty-strings-are-not-namespaces, cluster-scoped-ignores-a-namespace, unknown-kind-unchanged, same-name-different-group), and 3 wire-level tests asserting the actual request line for a namespaceless Secret, a caller-scoped ConfigMap, and a cluster-scoped kind.

### Test scope note

`validate` is proven at the wire (the HTTP request line is asserted against a recording stand-in). `apply` and `diff` are verified structurally — they construct the `Api` identically and route every scope decision through the exhaustively-tested `apply_namespace` (there is no remaining `match &r.namespace` in `manifest.rs`) — but not through their own HTTP test, which would need a kubeconfig-backed mock-server harness the repo doesn't have elsewhere. That harness can be added if reviewers want the two remaining paths wire-proven.

### DIFF
--- a/apps/desktop/src/components/EditResourceTab.tsx
+++ b/apps/desktop/src/components/EditResourceTab.tsx
@@ -46,6 +46,7 @@ export function EditResourceTab({
   return (
     <ManifestEditor
       context={context}
+      namespace={namespace ?? undefined}
       yaml={yaml}
       onYamlChange={setYaml}
       ariaLabel="Edit resource YAML"
--- a/apps/desktop/src/components/ManifestEditor.test.tsx
+++ b/apps/desktop/src/components/ManifestEditor.test.tsx
@@ -78,7 +78,7 @@ describe("ManifestEditor", () => {
     applyManifestMock.mockResolvedValue({ applied: true, documents: [{ kind: "ConfigMap", name: "web", applied: true }] });
     render(<Harness mode="create" />);
     fireEvent.click(screen.getByRole("button", { name: "Create" }));
-    await waitFor(() => expect(applyManifestMock).toHaveBeenCalledWith("kind-dev", expect.stringContaining("ConfigMap"), false));
+    await waitFor(() => expect(applyManifestMock).toHaveBeenCalledWith("kind-dev", expect.stringContaining("ConfigMap"), false, undefined));
     expect(notifyMock.success).toHaveBeenCalled();
   });
 
@@ -132,7 +132,7 @@ describe("ManifestEditor", () => {
     const force = await screen.findByRole("button", { name: /force apply/i });
     expect(screen.getByText(/kubectl/)).toBeDefined();
     fireEvent.click(force);
-    await waitFor(() => expect(applyManifestMock).toHaveBeenLastCalledWith("ctx", expect.any(String), true));
+    await waitFor(() => expect(applyManifestMock).toHaveBeenLastCalledWith("ctx", expect.any(String), true, undefined));
   });
 
   it("drawer (non-fill) surfaces conflicts too: Force apply appears and re-applies with force", async () => {
@@ -164,7 +164,7 @@ describe("ManifestEditor", () => {
     const force = await screen.findByRole("button", { name: /force apply/i });
     expect(screen.getByText(/kubectl/)).toBeDefined();
     fireEvent.click(force);
-    await waitFor(() => expect(applyManifestMock).toHaveBeenLastCalledWith("ctx", expect.any(String), true));
+    await waitFor(() => expect(applyManifestMock).toHaveBeenLastCalledWith("ctx", expect.any(String), true, undefined));
   });
 
   it("multi-doc conflict banner names each conflicting document, not the applied one", async () => {
--- a/apps/desktop/src/components/ManifestEditor.tsx
+++ b/apps/desktop/src/components/ManifestEditor.tsx
@@ -34,6 +34,7 @@ const CodeEditor = lazy(() => import("../ui/CodeEditor").then((m) => ({ default:
  */
 export function ManifestEditor({
   context,
+  namespace,
   yaml,
   onYamlChange,
   ariaLabel = "Manifest YAML",
@@ -48,6 +49,9 @@ export function ManifestEditor({
   onApplied,
 }: {
   context: string;
+  /** Namespace this editor is scoped to, applied to a document that names
+   *  none. Without it the backend falls back to `default` (#404). */
+  namespace?: string;
   yaml: string;
   onYamlChange: (yaml: string) => void;
   ariaLabel?: string;
@@ -170,7 +174,7 @@ export function ManifestEditor({
   async function doApply(force: boolean) {
     setBusy(true);
     setError("");
-    const out = await applyManifest(context, yaml, force);
+    const out = await applyManifest(context, yaml, force, namespace);
     setBusy(false);
     if (out.error) {
       setError(describeError(out.error).detail);
@@ -241,7 +245,7 @@ export function ManifestEditor({
     let active = true;
     setDiffing(true);
     const t = setTimeout(() => {
-      void diffManifest(context, yaml).then((out) => {
+      void diffManifest(context, yaml, namespace).then((out) => {
         if (!active) return;
         const docs = out.error ? [] : out.documents ?? [];
         // After a successful apply, adopt the freshly-bumped resourceVersion as
@@ -274,7 +278,7 @@ export function ManifestEditor({
         minHeight={fill ? undefined : 320}
         maxHeight={fill ? undefined : 520}
         schemaValidate={(y) =>
-          validateManifest(context, y).then((r) => (r.valid === false ? r.errors ?? [] : []))
+          validateManifest(context, y, namespace).then((r) => (r.valid === false ? r.errors ?? [] : []))
         }
         schemaSource={(apiVersion, kind) =>
           openApiSchema(context, apiVersion, kind).then((r) => ("error" in r ? null : r))
--- a/apps/desktop/src/components/NewResourceEditor.test.tsx
+++ b/apps/desktop/src/components/NewResourceEditor.test.tsx
@@ -32,7 +32,7 @@ describe("NewResourceEditor", () => {
 
     fireEvent.click(screen.getByRole("button", { name: "Create" }));
 
-    await waitFor(() => expect(applyManifestMock).toHaveBeenCalledWith("kind-dev", editor.value, false));
+    await waitFor(() => expect(applyManifestMock).toHaveBeenCalledWith("kind-dev", editor.value, false, "prod"));
     expect(await screen.findByText(/Applied Service/)).toBeDefined();
     expect(onCreated).toHaveBeenCalled();
     // Editor is still present (tab stays open to create more).
--- a/apps/desktop/src/components/NewResourceEditor.tsx
+++ b/apps/desktop/src/components/NewResourceEditor.tsx
@@ -119,6 +119,7 @@ export function NewResourceEditor({
   return (
     <ManifestEditor
       context={context}
+      namespace={ns}
       yaml={yaml}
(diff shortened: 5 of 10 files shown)


deveshk0:
The diagnosis is right and the fix is in the right place. Scope belongs to the kind, the read paths already knew that, and putting one helper in front of all three write paths — rather than patching apply and leaving validate and diff to drift — is the version of this I wanted. `r.namespace` only feeds `apply_namespace` now, so there's no second opinion left in the file.

Two calls worth singling out because they're easy to get wrong:

- Leaving unknown kinds alone (`manifest.rs:510`). Defaulting a CRD to namespaced would have traded a known bug for a silent one, and the reasoning is written down where the next person will find it.
- Gating `gvk_for` on group and version. A custom `Service` in someone else's group getting forced into a namespace is exactly the sort of thing that surfaces six months later with nobody able to explain it.

No blockers. Smaller things:

- `manifest.ts:318,339,377` — `namespace` goes in ahead of `invoke` in all three, so the tests now read `applyManifest("ctx", "kind: ConfigMap", undefined, undefined, invoke)`. Type-safe, but that's four positionals, two optional, with the seam last. Either move `namespace` after `invoke` or take an options object. Not blocking — I'd rather this landed than got refactored first.
- `manifest.rs:579,672,918` — the three-line `match` that builds the `Api` is still copied three times. Fold that in too (return the `Api`, or an `api_for(client, &ar, ns)` next to it) and the "apply and diff are verified structurally" note in your description stops being a note: one construction site instead of three that agree today.
- `manifest.rs:910` — `doc_ns` is the resolved namespace, not the document's. The comment says so, the name says the opposite. `applied_ns`, or just `ns`.
- `manifest.rs:996` — `recording_server` takes one `read` and assumes the request line landed in it. It will, nearly always. But `:1052` and `:1068` index `lines[0]` with no emptiness check, so on the day it doesn't you get an index panic instead of your own message. `:1028` guards it; the other two should match.
- `apply_namespace` is called by validate and diff as much as by apply. `namespace_for` reads truer.

One question, not for this PR: the last resort is a hardcoded `default`, which matches the read paths, so I'm fine with it here. But `kubectl` uses the context's namespace, and a context with `namespace: team-a` set gets `default` from us — including over MCP, where nothing threads a tab namespace in. Worth an issue?

Same for CRDs — a namespaced custom resource applied without a namespace still reproduces #404. Right call to leave it, but let's put that on the issue before closing it so it doesn't come back as a fresh report.

I checked `capability-catalog.json` and `docs/mcp-catalog.md` since `ApplyIn` grew a field — neither records input schemas, so there's nothing to regenerate.

Line anchors are from the head of `fix/404-apply-namespace-scope`. I read this on the branch but didn't build or run it — the 5464 / 92.89% figures are as reported.

deveshk0 (approved):

PR #421, opened 2026-09-01
fix(copy): add shared inline clipboard feedback

### LINKED ISSUE #410: "Copy as kubectl" gives no confirmation — its success toast reaches nobody in the new design

saiyam1814:
## Summary

- introduces one shared clipboard controller with a 1.4-second confirmation window, repeat-click restart, latest-write ordering, and unmount cleanup
- provides visible success/failure states plus a polite assistive-technology announcement; failed writes never claim success
- applies the same behavior to CopyCommand, KubectlPreview, resource row/detail actions, Overview, and forwarded addresses
- preserves reduced-motion behavior and removes reliance on the absent new-design toast host

## Verification

- focused clipboard race/timer/accessibility tests and regression tests for every consuming surface
- workspace TypeScript typecheck
- complete frontend suite: 5,478/5,478 tests passed with coverage

Closes #410

### DIFF
--- a/packages/ui-kit/src/ActionBar.test.tsx
+++ b/packages/ui-kit/src/ActionBar.test.tsx
@@ -155,6 +155,16 @@ describe("ActionBar", () => {
       expect(screen.queryByRole("dialog")).toBeNull();
     });
 
+    it("can keep a feedback-bearing action visible after it runs", async () => {
+      setup({
+        max: 1,
+        actions: [action("Logs"), action("Copy", { closeOnSelect: false })],
+      });
+      await openMenu();
+      await userEvent.click(screen.getByRole("button", { name: "Copy" }));
+      expect(screen.getByRole("dialog")).toBeDefined();
+    });
+
     it("offers no menu when everything fits", () => {
       setup({ max: 4 });
       expect(screen.queryByRole("button", { name: "More actions" })).toBeNull();
--- a/packages/ui-kit/src/ActionBar.tsx
+++ b/packages/ui-kit/src/ActionBar.tsx
@@ -19,6 +19,8 @@ export interface ActionBarAction {
    * a blocked action with no explanation is worse than no action at all.
    */
   disabledReason?: string;
+  /** Keep an overflow row open while its transient result replaces the label. */
+  closeOnSelect?: boolean;
   onSelect: () => void;
 }
 
@@ -176,7 +178,7 @@ export function ActionBar({
                       // Left open on purpose: a menu that shuts looks like the
                       // action was taken.
                       if (blocked) return;
-                      close();
+                      if (a.closeOnSelect !== false) close();
                       a.onSelect();
                     }}
                   >
--- a/packages/ui-kit/src/ContextMenu.test.tsx
+++ b/packages/ui-kit/src/ContextMenu.test.tsx
@@ -90,6 +90,14 @@ describe("ContextMenu", () => {
     await waitFor(() => expect(screen.queryByRole("menu")).toBeNull());
   });
 
+  it("can keep a feedback-bearing item visible after it is picked", async () => {
+    const onPick = vi.fn();
+    await open({ items: [{ label: "Copy", closeOnPick: false, onPick }] });
+    fireEvent.click(screen.getByRole("menuitem", { name: "Copy" }));
+    expect(onPick).toHaveBeenCalledTimes(1);
+    expect(screen.getByRole("menu")).toBeDefined();
+  });
+
   it("draws a separator that is not itself an item", async () => {
     await open();
     const menu = screen.getByRole("menu");
--- a/packages/ui-kit/src/ContextMenu.tsx
+++ b/packages/ui-kit/src/ContextMenu.tsx
@@ -19,6 +19,8 @@ export type ContextMenuItem =
       hint?: string;
       /** Tints the row with the danger colour, over a label that already says so. */
       danger?: boolean;
+      /** Keep the menu open while transient feedback replaces this row's label. */
+      closeOnPick?: boolean;
       onPick: () => void;
     };
 
@@ -118,7 +120,10 @@ export function ContextMenu({ items, children, label, onOpenChange }: ContextMen
                 // what the item does nor what a speech-input user would say to
                 // reach it. (#320)
                 aria-label={item.label}
-                onSelect={item.onPick}
+                onSelect={(event) => {
+                  if (item.closeOnPick === false) event.preventDefault();
+                  item.onPick();
+                }}
               >
                 {/* Always taken, filled or not: the design lines the labels up
                     in a column, and an icon present on only some items shoves
--- a/packages/ui-kit/src/CopyCommand.test.tsx
+++ b/packages/ui-kit/src/CopyCommand.test.tsx
@@ -61,7 +61,7 @@ describe("CopyCommand", () => {
     expect(container.querySelector(".copy-command-check")).not.toBeNull();
   });
 
-  it("leaves the command readable on a machine with no clipboard", async () => {
+  it("keeps the command readable and reports a clipboard failure", async () => {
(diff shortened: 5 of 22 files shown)


saiyam1814:
Closing as superseded by #413, which merged the same #410 fix after review. Keeping this implementation would duplicate the now-shipped useCopied/CopyIconButton primitives and create two competing clipboard-feedback architectures.

PR #431, opened 2026-09-03
fix(ui): display node memory capacity in human-readable units

### LINKED ISSUE #430: bug(ui): node detail shows raw Kibibyte memory instead of a human-readable unit
before:
Image

after:
Image

 
- Node detail capacity section displays memory as the raw k8s value: e.g. 32242636Ki, 32345036Ki 
- It should be more readable if converting it to Gibibyte unit like 30.7Gi / 30.8Gi.

mznet:
fix #430


deveshk0 (approved):
LGTM

PR #456, opened 2026-09-07
fix(docker): install the pnpm version the lockfile was written by

### LINKED ISSUE #455: Docker release build fails: image runs pnpm 9, lockfile needs 10+
Both `docker image` jobs have failed on every release run since the CVE pins moved into `pnpm-workspace.yaml`, at `RUN pnpm install --frozen-lockfile`:

```
ERR_PNPM_LOCKFILE_CONFIG_MISMATCH  Cannot proceed with the frozen installation.
The current "overrides" configuration doesn't match the value found in the lockfile
```

Seen in runs 34057263143 and 34119435556 (amd64 and arm64 alike).

## Cause

The `overrides:` block lives in `pnpm-workspace.yaml`, which is where pnpm 10+ reads it from. The Dockerfile installs `pnpm@9`, and pnpm 9 only reads overrides from `package.json` under `pnpm.overrides`. So inside the image pnpm sees no overrides at all, while the lockfile records nine of them, and a frozen install refuses to continue.

CI does not hit this: `pnpm/action-setup` reads `packageManager` from `package.json`, which pins `pnpm@11.24.0`. The Dockerfile is a second, stale source of truth for the same thing. Its comment even claims the pinned major is what wrote the lockfile, which stopped being true when local moved to 11.

## Fix

Derive the version in the image from `packageManager`, the same field CI reads, so there is one source of truth and no version to forget to bump.

deveshk0:
Closes #455.

Both `docker image` jobs have failed on every release run since the CVE pins moved into `pnpm-workspace.yaml` (runs 34057263143 and 34119435556, amd64 and arm64 alike), at `pnpm install --frozen-lockfile`:

```
ERR_PNPM_LOCKFILE_CONFIG_MISMATCH  Cannot proceed with the frozen installation.
The current "overrides" configuration doesn't match the value found in the lockfile
```

## Why

The image installed `pnpm@9`, and pnpm 9 reads overrides only from `package.json` under `pnpm.overrides`. The overrides live in `pnpm-workspace.yaml`, which is where pnpm 10 and later look. Inside the image there were therefore no overrides at all while the lockfile recorded nine, and a frozen install will not continue past that mismatch.

CI never hit it: `pnpm/action-setup` reads `packageManager` from `package.json`, which pins `pnpm@11.24.0`. The Dockerfile was a second, stale copy of the same fact, and its comment still claimed the pinned major was what wrote the lockfile.

## What changed

The image now derives the version from `packageManager`, the same field CI reads, so there is one place to change it:

```dockerfile
RUN npm install -g "pnpm@$(node -p "require('./package.json').packageManager.split('@').pop()")"
```

`package.json` is copied on its own, ahead of the lockfile, so the install layer stays cached when only the lockfile moves.

The `.dockerignore` change is what makes this checkable without cutting a release. A bare `node_modules` matches the root one only, so a developer's `packages/ui-next/node_modules` was copied in on top of the install the image had just done. Its workspace links are absolute host paths and dangle inside the container, so the build failed to resolve `@srelens/ui-kit/gallery` on any machine that had run pnpm install, and never in CI. Now every `node_modules` is excluded at any depth.

`AGENTS.md` records the trap: the version has one home, and the image is buildable locally.

## Verified locally

`docker build --target frontend .` installs pnpm 11.24.0, passes the frozen install, and vite writes `apps/desktop/dist`. A full `docker build` is running and I will report it here.

Note the release run that exposed this is otherwise healthy: all seven TUI targets built, and the macOS binaries came back from the notary service Accepted.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,11 @@
 target
-node_modules
-apps/desktop/node_modules
+# Every node_modules, at any depth. A bare `node_modules` matches the root one
+# only, so a developer's `packages/ui-next/node_modules` was copied in on top
+# of the install the image had just done -- and its workspace links are
+# absolute host paths, which dangle inside the container. The build then could
+# not resolve `@srelens/ui-kit/gallery`, on a machine that had run pnpm
+# install and never in CI.
+**/node_modules
 apps/desktop/dist
 apps/desktop/src-tauri/target
 .git
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,3 +219,29 @@ jobs:
           name: smoke-failure-artifacts
           path: ${{ runner.temp }}/smoke-artifacts
           if-no-files-found: ignore
+
+  docker:
+    name: docker image
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Set up Buildx
+        uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
+      # Until now nothing but the release workflow ever built this file, and
+      # two dev releases shipped no container image before anyone noticed. Two
+      # separate breaks were waiting in it, one per stage: a pnpm version the
+      # lockfile had outgrown, and a workspace member the backend stage never
+      # copied. Building only the frontend stage here would have caught one of
+      # them.
+      #
+      # So the whole image, which makes this the slowest job in CI on any
+      # change to crates/. That is the price of the image being built at all
+      # before a release depends on it. The layer cache carries the Rust stage
+      # across runs that do not touch it.
+      - name: Build the image
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          push: false
+          cache-from: type=gha,scope=docker-ci
+          cache-to: type=gha,mode=max,scope=docker-ci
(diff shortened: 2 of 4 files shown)


deveshk0:
Second break found while verifying the first, and folded into this PR.

With the pnpm version fixed the image build reached its next failure, in the backend stage:

```
error: failed to load manifest for workspace member `/src/apps/tui`
  failed to read `/src/apps/tui/Cargo.toml`
```

`apps/tui` joined the workspace when the TUI landed. The backend stage copies `crates` and `apps/desktop/src-tauri` only, and cargo loads every member's manifest even when building a single package. Nothing in the image builds or ships the TUI, so it just needs the member to be there.

That also settles what the new CI job should build. One break was in each stage, so building only the frontend stage would have caught one of them. The job builds the whole image. On a change to `crates/` it becomes the slowest job in CI, which is the price of knowing the image builds before a release depends on it.

A local `docker build` is running against the branch and is past manifest loading into the Rust compile. I will report the result.

navneet066 (approved):

deveshk0:
Verified, both in CI and locally.

CI is green, including the new job, which built the whole image on a runner:

| job | |
| --- | --- |
| docker image | success |
| backend | success |
| frontend | success |
| smoke (webdriver) | success |

Locally, `docker build .` completes through both stages, and the image runs:

```
healthz: ok
kubectl v1.36.3
helm v3.21.3+g1ad6e68
uid=10001(srelens) gid=10001(srelens)
```

It serves the embedded bundle at `/`, so the frontend stage's output really did reach the binary. Started without configuration it exits 1 with `web mode requires auth`, and with auth but no key it exits 1 asking for `SRELENS_MASTER_KEY` — both intended, and both reached only because the image now builds at all.

For the record on the release this came from: v0.8.1-153 went out complete apart from the container image. All seven TUI archives, the checksum file, and a GPG signature on each.

PR #457, opened 2026-09-07
release: promote dev to main — the TUI, the new design, and a container image again

deveshk0:
112 commits since v0.8.0.

## What lands

**A terminal client.** `srelens-tui` is new in this window and arrived nearly whole: cluster overview with capacity gauges and GPU allocation, a workloads screen, a live events stream with warning triage, deep node and Helm inspectors, port-forwarding, an AI assistant with per-cluster state and rich markdown, deep links, mouse selection and OSC 52 clipboard support across every view. It ships as a signed binary for seven targets, updates itself on a stable or dev channel, and installs with `brew install srelens/tap/srelens-tui`.

**The new design.** Steps 1 through 9 of the ui-next migration, including the agent screen and console dock as two views of one run, plus the Topology screen laid out by flow with observed traffic.

**Cluster depth.** A node and GPU hardware inspector with live allocations, dynamic printer columns giving kubectl parity for custom resources, and a generic `k8s.queryPodEndpoint` tool that reaches a pod's `/metrics` or any HTTP endpoint through API port-forwarding.

**Fixes worth naming.** Apply scope resolved from the kind rather than the document, list AGE derived from a timestamp so it ticks, the resource relationship tree nesting Deployment then ReplicaSet then Pod, the macOS re-sign that was letting dyld SIGKILL a freshly installed binary, and the window commands every close path goes through.

**Supply chain.** The CVE pins moved to where pnpm 11 reads them, pnpm itself pinned, seven advisories cleared across fast-uri, qs and browserslist, and the three base images bumped.

## Merging this cuts a stable release

Worth knowing what that run does that a dev run never has:

- It is the first stable release to carry TUI archives, so the `homebrew` job runs for real for the first time. The tap is stable-only by design, so this is the first push to `srelens/homebrew-tap` and the first time `brew install srelens/tap/srelens-tui` resolves.
- `size-baseline` and `publish-release` are main-only and will run.
- Apple signing is required rather than optional on stable. It is verified working — the dev run notarized all four macOS binaries and came back Accepted.

## One thing to wait for

The container image has been missing from every dev release since the CVE pins moved, two separate breaks, both fixed on dev an hour ago. Release run 34122522872 is in flight and is the first to exercise those fixes end to end. Its two `docker image` legs are the thing to see green before this merges; everything else in it has already passed once.


deveshk0:
The thing to wait for has cleared. Release run 34122522872 finished green, both docker legs included:

```
success  docker image (amd64)
success  docker image (arm64)
success  docker manifest
```

`ghcr.io/srelens/srelens:0.8.1-154` is published as a multi-arch index covering `linux/amd64` and `linux/arm64` — the first dev pre-release to carry a container image since the CVE pins moved. `size-baseline`, `publish-release`, `aur` and `homebrew` skipped, all main-only by design, and they are exactly what merging this will exercise for the first time.

Nothing left blocking from the release side.

